PR #17, opened 2016-08-16
Install Devise

hortongn:


jamesvanmil:
Looks good - one thing I noticed is that the specs are running in the same order each time you run `rspec`. I'll open a new issue to randomize them with a seed.

hortongn:
Good catch on randomizing the specs

PR #22, opened 2016-08-18
Feature/#3 create workgroup

### LINKED ISSUE #3: Create a workgroup
User story: https://trello.com/c/b6TkhJRW/17-creating-a-workgroup

Depends on #2

jamesvanmil:
Some notes:

It's apparently challenging to test routes and requests for authenticated resources, because it's not straightforward to stub the authentication checks when doing full the stack testing that these test types perform. There's some workarounds, but they didn't work out of the box for me: https://github.com/plataformatec/devise/issues/1670

Okay to write off routing tests and request test from the get-go and rely on our integration and feature tests to touch all these pages?

More notes:

A lot of the controller tests are waiting around for attributes to be defined. We don't have any yet, so I set these up to remain pending.

Final notes:

There's probably more that could be done with view tests, but nothing that we wouldn't be undoing once our roles and permissions are in place.

Closes #3


jamesvanmil:
rebasing against develop for review

hortongn:
After I have create a new workgroup or edit a workgroup, the resulting show page lists the "Workgroup was successfully created/edited" twice.  The workgroup show view includes a `<%= notice %>`, but we already have that on the page in our layout view from when Devise was added:
https://github.com/uclibs/ucrate/blob/develop/app/views/layouts/application.html.erb#L14

So I think it's safe to remove the notice from the workgroup show view: https://github.com/uclibs/ucrate/blob/develop/app/views/layouts/application.html.erb#L14

PR #24, opened 2016-08-22
add title and description to workgroups and update specs

### LINKED ISSUE #7: Add title and description to workgroups
Expose the date and time the workgroup was created

User story: https://trello.com/c/60L1VDHS/28-basic-workgroup-metadata

Depends on #3

lawhorkl:
closes #7

### DIFF
--- a/app/controllers/workgroups_controller.rb
+++ b/app/controllers/workgroups_controller.rb
@@ -26,7 +26,6 @@ class WorkgroupsController < ApplicationController
   # POST /workgroups.json
   def create
     @workgroup = Workgroup.new(workgroup_params)
-
     respond_to do |format|
       if @workgroup.save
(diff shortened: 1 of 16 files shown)


hortongn:
I think we need to fix the way the migration is handled.  You're adding two columns to an existing migration instead of creating a new one.  I can show you the problem I'm having tomorrow.

hortongn:
The title and description migration is working for me after your change.  

One more issue: you have a "destroy" link in your workgroup index view, but there's no destroy action action in the controller.  So the link fails.  Just remove the destroy link from the view.  Deleting a workgroup is beyond the scope of this issue anyway.  After that I think we're good to merge.

lawhorkl:
Hm... That's odd, I must have not caught it when I copied the generated views.

lawhorkl:
@hortongn removed the line!

PR #41, opened 2016-08-28
Add users and roles to workgroups

### LINKED ISSUE #4: Add/edit/delete users from an existing workgroup
Add, edit, or delete users from a workgroup and select a role for each user.

Roles: owner, editor, reader

User story: https://trello.com/c/RSvT9m8r/31-add-authorized-users-to-workgroup

Depends on #3

hortongn:
closes #4

### DIFF
--- a/app/assets/javascripts/workgroup_membership.coffee
+++ b/app/assets/javascripts/workgroup_membership.coffee
@@ -1,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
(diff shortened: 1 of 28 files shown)


hortongn:
I temporarily merged this branch locally with the simplecov branch and verified we still have 100% test coverage.

jamesvanmil:
Still to look at specs. Functionality checks out, but I have a few thoughts:

* I'm wondering if we should have [nested resources](http://guides.rubyonrails.org/routing.html#nested-resources) for the routes - I think this would mean that we go to `/workgroup/1/memberships/new` instead of to `workgroup_memberships/new?id=1`. Still trying to pin down the advantage of that, but something about it feels better.
* What do you think about combining the user management for and the role management forms? If `workgroup_memberships/new` displayed a list of users with with radio buttons for the roles next to each, the interactions for system users could be more direct. (This might be something to iterate on with the POs)

PR #97, opened 2018-02-20
Increase default 500MiB limit to 3GiB

### LINKED ISSUE #75: Limit the size of files that can be uploaded from user's computer
### Descriptive summary

Set a 3GB limit (like in Scholar@UC) for the maximum size of a file a user can upload.  The limit only applies to local uploads (not cloud files).

Similar change in Scholar@UC: https://github.com/uclibs/scholar_uc/blob/develop/app/assets/javascripts/hyrax/uploader.js#L32

Trello: https://trello.com/c/0sHAiBex/98-limit-upload-file-sizes

jayanthdeejay:
Fixes #75  

Override JQuery file upload validation script to increase the file size limit from 500MiB to 3GiB per uploaded file.

Tested it locally with various file sizes. Throws "File is too large" for files larger than 3GiB.

### DIFF
--- a/vendor/assets/javascripts/fileupload/jquery.fileupload-validate.js
+++ b/vendor/assets/javascripts/fileupload/jquery.fileupload-validate.js
@@ -1,0 +1,122 @@
+/*
+ * jQuery File Upload Validation Plugin
+ * https://github.com/blueimp/jQuery-File-Upload
+ *
+ * Copyright 2013, Sebastian Tschan
+ * https://blueimp.net
+ *
+ * Licensed under the MIT license:
+ * http://www.opensource.org/licenses/MIT
+ */
+
+/* global define, require, window */
+
+(function (factory) {
+    'use strict';
+    if (typeof define === 'function' && define.amd) {
+        // Register as an anonymous AMD module:
+        define([
+            'jquery',
+            './jquery.fileupload-process'
+        ], factory);
+    } else if (typeof exports === 'object') {
+        // Node/CommonJS:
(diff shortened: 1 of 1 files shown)


hortongn:
@jayanthdeejay I just found this https://github.com/uclibs/ucrate/blob/develop/config/initializers/hyrax.rb#L37

Can you see if setting the maxfilesize in the Hyrax initializer works without having to edit `jquery.fileupload-validate.js`?  That would be a cleaner solution.

jayanthdeejay:
@hortongn I did uncomment the respective lines yesterday. For some reason, I could not upload any files after doing so. I will once again look into it.

vendor/assets/javascripts/fileupload/jquery.fileupload-validate.js
```diff
@@ -0,0 +1,122 @@
...
+
+            // Function returning the current number of files,
+            // has to be overriden for maxNumberOfFiles validation:
+            getNumberOfFiles: $.noop,
+
+            // Error and info messages:
+            messages: {
+                maxNumberOfFiles: 'Maximum number of files exceeded',
+                acceptFileTypes: 'File type not allowed',
+                maxFileSize: 'File is too large',
```
hortongn: @jayanthdeejay I confirmed the 3GB size limit works.  However, can you change this line 75 to `maxFileSize: 'This file is too large to upload via the website. For files larger than 3 gigabytes, upload them using the cloud providers option below or use the Contact Us form to request other ways to add your content.'`

PR #102, opened 2018-02-22
Adds name in user.rb.

### LINKED ISSUE #100: Use the user's name when it exists
### Descriptive summary

Change the application to use first and last name instead of email address when the user has entered their name into their profile.

https://github.com/uclibs/scholar_uc/commit/68f02b78b6b46602546ccdac0309dc33658dd358

Refs: 58

changy66:
Fixes #100

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,4 +24,9 @@ class User < ApplicationRecord
   def to_s
     email
   end
+
+  def name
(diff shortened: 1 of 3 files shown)


hortongn:
@changy66 Can you please add a User model spec for your #name method?  It should test that the #name method returns the correct value when the user has a name and when the user does not have a name.  I can work with you on this if you like.

PR #105, opened 2018-02-23
Add recpatcha to contact form for non-authenticated users

### LINKED ISSUE #61: Add captcha to contact form
### Descriptive summary

Add a Google captcha to the contact form to prevent spam.  The captcha should only be active for non-authenticated users.  Implement this the same way as Scholar@UC.

Scholar@UC commit:
https://github.com/uclibs/scholar_uc/commit/6ca1decdae8205f9142b4bb40a638053194b5289

Trello: https://trello.com/c/GBy3p60n/96-contact-form-captcha

jayanthdeejay:
Fixes #61;

May need to verify that [this](https://github.com/uclibs/ucrate/blob/f94f3b270b0a7ac0c6ec78ea45d8c93499db698b/app/controllers/hyrax/contact_form_controller.rb#L38) indeed receives the secret key and recaptcha authentication works. But, apart from that, this PR is in good shape.

* Adds Google reCaptcha to the contact form. reCaptcha verification only for unauthenticated users.

### DIFF
--- a/spec/controllers/hyrax/contact_form_controller_spec.rb
+++ b/spec/controllers/hyrax/contact_form_controller_spec.rb
@@ -1,0 +1,117 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Hyrax::ContactFormController do
+  routes { Hyrax::Engine.routes }
+  let(:user) { create(:user) }
+  let(:required_params) do
+    {
+      category: "Depositing content",
+      name: "Gandalf the Grey",
+      email: "gandalf@middle.earth",
+      subject: "When in doubt,",
+      message: "Follow your nose!"
+    }
+  end
+
+  describe 'while user is unauthenticated' do
+    it 'successfully allows reCaptcha' do
+      described_class.any_instance.stub(:verify_google_recaptcha).and_return(true)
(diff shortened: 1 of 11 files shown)


spec/controllers/hyrax/contact_form_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  describe 'while user is unauthenticated' do
+    it 'successfully allows reCaptcha' do
+      described_class.any_instance.stub(:verify_google_recaptcha).and_return(true)
+      Hyrax::ContactMailer.any_instance.stub(:mail).and_return(true)
+      post :create, params: { contact_form: required_params }
+      expect(flash[:notice]).to match(/Thank you for your message/)
+    end
+
+    it 'fails on reCaptcha failure' do
+      described_class.any_instance.stub(:passes_captcha_or_is_logged_in?).and_return(false)
```
hortongn: @jayanthdeejay I think you are missing coverage in the `verify_google_recaptcha` method because your spec is stubbing `passes_captcha_or_is_logged_in?` to `false`.  So the spec never runs the code in `passes_captcha_or_is_logged_in?`

Try removing the stub at line 27 and then I think you will have better coverage.

PR #119, opened 2018-03-06
Add the view for change password

### LINKED ISSUE #113: Allow users to change their account password (when not logged in via Shibboleth)
### Descriptive summary

Users should be able to change their account password.  When a user is editing their profile, there should be a section of the form where they can enter a new password (and verify it by entering the new password a second time).

Devise has the built in ability to do this, but the edit profile form doesn't currently have those fields.

We only want users to have the change password option when they are *NOT* logged in via Shibboleth.  I believe that is how we do it in Scholar.

How they did it in Hyku: https://github.com/samvera-labs/hyku/commit/08e2a4daf2ad565e2a35d01f957f9c0f0482062c

changy66:
Fixes #113

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -16,6 +16,7 @@
         <li><%= link_to t("hyrax.toolbar.dashboard.menu"), hyrax.dashboard_path %></li>
 
         <li class="divider"></li>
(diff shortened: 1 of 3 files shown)


hortongn:
@changy66 Good job.  I'm going to merge this, but I have a couple more changes we should make.  Let's spend some time Wednesday morning to talk about this issue and your pull request #118

PR #138, opened 2018-03-21
Hide the lease visibility option

### LINKED ISSUE #128: Hide the lease visibility option
### Descriptive summary

When creating or editing a work, users should *not* see the option to set a lease.

Similar commit in Scholar: https://github.com/uclibs/scholar_uc/commit/d3e3c0dcfc37be3dc9fc9f151600c08e2ef7e8f5

Trello: https://trello.com/c/JtmzJOSX/101-hide-lease-visibility-option

changy66:
Fixes #128


scherztc:
I know that we are only hiding the lease option from the view here.

But the functionality is still viewable from the administrative dashboard: http://localhost:3000/leases?locale=en

screen shot 2018-03-23 at 12 48 00 pm

I still approve but @hortongn needs to make new issues if we want those things removed from the admin dashboard.

scherztc:
@changy66 :  We have to deal with the Permissions on the EDIT file partial, too.  Maybe a test for that.

screen shot 2018-03-23 at 3 35 18 pm

PR #144, opened 2018-03-27
Feature/#78c collection export

### LINKED ISSUE #78: Export metadata from a collection
### Descriptive summary

A user should be able to select a collection and export the metadata for all works in that collection to a CSV file that can be downloaded to their local machine.

Trello: https://trello.com/c/WwAQ7DJK/66-export-metadata

Depends on #108

jamesvanmil:
Closes #78 

Adds MVC for Collection Metadata Export

* 2 Hyrax overrides to add buttons/links to the new feature:
  * `app/views/hyrax/dashboard/collections/_show_actions.html.erb`
  * `app/views/hyrax/my/_collection_action_menu.html.erb`
  * `spec/views/hyrax/dashboard/collections/_show_actions.html.erb_spec.rb`
  * `spec/views/hyrax/my/_collection_action_menu.html.erb_spec.rb`
* Permissions for `collection_export` inherit from associated collection - if that is missing, it looks at the user who created the instance
* Specs for:
  * Controller
  * Feature
  * Ability
* Added static 403 error page

Considerations:

* There's a blocker with the feature spec - see the commented section - this feature is well tested via the view override specs (which look for the button) and  other feature spec expectations, which are looking at the same actions.
* No model code/specs - there was no need for validations for the model, so I didn't add anything here
* The need for view specs is covered by the feature spec.


jamesvanmil:
I'm getting fairly consistent failures from Travis on this:

```
rspec ./spec/features/uc_shibboleth_spec.rb:62 # UC account workflow overridden devise sign-in page shows a shibboleth login link if shibboleth is enabled
```

I don't think this is related to my PR - environment issue or flaky test, maybe?

app/views/hyrax/dashboard/collections/_show_actions.html.erb
```diff
@@ -0,0 +1,36 @@
...
+                  data: { nestable: presenter.collection_type_is_nestable?,
+                          hasaccess: true } do %>
+                  <%= t('hyrax.collection.actions.nested_subcollection.button_label') %>
+                <% end %>
+<% end %>
+
+<%= link_to main_app.collection_exports_path(collection_id: presenter.id),
+            method: 'post',
+            class: 'btn btn-primary export-collection-button',
+            title: t("hyrax.collection.actions.export_collection") do %>
```
hortongn: @jamesvanmil You are missing an entry in config/locales/hyrax.en.yml for `hyrax.collection.actions.export_collection`

hortongn:
@jamesvanmil I'm not necessarily asking that it go into this PR, but do you have anything linking to the Collection Exports page (`/collection_exports`)?  The user ends up on that page when they create an export, but what if they just want to see their existing exports?  How do they get there?  Maybe we need a link to `/collection_exports` on the dashboard side bar?  I'm happy with that being a new issue.

app/views/collection_exports/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  </thead>
+
+  <tbody>
+    <% @collection_exports.each do |collection_export| %>
+      <tr scope="row">
+        <td><%= collection_export.collection_id %></td>
+        <td><%= collection_export.created_at %></td>
+        <td><%= link_to 'Download File', collection_export_download_path(collection_export) %></td>
+        <% if can? :destroy, collection_export %>
+          <td><%= link_to 'Delete', collection_export, method: :delete, data: { confirm: 'Are you sure?' } %></td>
```
hortongn: @jamesvanmil Instead of just saying "Are you sure?" in the pop-up modal, we should be consistent with what the user sees when they delete works, etc.  Maybe something like "Click OK to delete this export or Cancel to cancel this operation."  Use i18n.

app/controllers/collection_exports_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  # POST /collection_exports
+  def create
+    if can? :show, Collection.find(collection_id)
+      @collection_export = CollectionExport.create(
+        export_file: new_collection_export_file.read,
+        collection_id: collection_id,
+        user: current_user.email
+      )
+
+      redirect_to collection_exports_url, notice: 'Collection export was successfully created.'
```
hortongn: @jamesvanmil Use i18n

app/controllers/collection_exports_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+
+  # DELETE /collection_exports/1
+  def destroy
+    if can? :destroy, @collection_export
+      @collection_export.destroy
+
+      respond_to do |format|
+        format.html { redirect_to collection_exports_url, notice: 'Collection export was successfully destroyed.' }
```
hortongn: @jamesvanmil Can we day "deleted" instead of "destroyed"?  Use i18n

app/views/collection_exports/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <th scope="col">Timestamp</th>
+      <th scope="col">File</th>
+      <th scope="col">Delete</th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% @collection_exports.each do |collection_export| %>
+      <tr scope="row">
+        <td><%= collection_export.collection_id %></td>
```
hortongn: @jamesvanmil This can be a future enhancement instead of this PR, but I think it would be nice to display the current collection title instead of just the ID.  If a user has many collection exports listed, they will have a hard time figuring out which collection the exports belong to if all they can see is the pid.

hortongn:
@jamesvanmil I don't think I like that a user can create an export of an empty collection.  It produces just an empty CSV file.  Maybe the 'Export Collection' button shouldn't display when a collection has no works?

hortongn:
@jamesvanmil Again, this can be a future PR, but the .csv files appear to get stored in tmp/ and that will be a problem in a production environment.  We should probably make that path configurable so it can stay tmp/ in development, but point to a shared drive in our load-balanced environment.

app/models/ability.rb
```diff
@@ -17,6 +17,17 @@ def custom_permissions
     #   can [:create], ActiveFedora::Base
     # end
 
+    can :show, CollectionExport do |collection_export|
+      collection_export.user == current_user.email ||
+        (current_user.can? :show, Collection.find(collection_export.collection_id))
```
hortongn: @jamesvanmil This line (22) throws an `Ldp::Gone in CollectionExportsController#index` error if `collection_export.collection_id` is from a deleted collection.  We only hit this condition when a collection export (from a since deleted collection) is owned by someone other than the `current_user`.

jamesvanmil:
Making a list of everything to address

- [ ] Checking ability when a collection is deleted
- [ ] Link to collection exports page
- [ ] Prevent export of an empty collection
- [ ] Implement I18n for collection export views
- [ ] Missing locale: `hyrax.collection.actions.export_collection`

jamesvanmil:
@hortongn I added the CSVs to the database: https://github.com/uclibs/ucrate/pull/144/files#diff-bb18e9268e82209adb21af10cc16906dR6

They're stored on `tmp` when produced, then persisted in the database when the `CollectionExport` is created.

hortongn:
@jamesvanmil These are the specs that are breaking with your latest commit:

```
rspec ./spec/features/collection_export_spec.rb:30 # collection export logged in collections that the user created do all the things we expect

rspec ./spec/views/hyrax/dashboard/collections/_show_actions.html.erb_spec.rb:44 # hyrax/dashboard/collections/_show_actions.html.erb when user cannot edit the document renders export collection link

rspec ./spec/views/hyrax/dashboard/collections/_show_actions.html.erb_spec.rb:31 # hyrax/dashboard/collections/_show_actions.html.erb when user can edit the document renders export collection link
```

hortongn:
@jamesvanmil I'm getting the error below when I try to do an export on a collection that has a sub-collection.  I had a parent collection that contained 1 work & 1  sub-collection.  The sub-collection contained just 1 work.

error

hortongn:
@jamesvanmil This is not crashing the app when I do an export, but if I export a collection that has a work with child works, the child works are not listed in the export file.

It's not a blocker for this PR, but I think we need a new issue to get collection exports to include child works.

jamesvanmil:
It doesn't look like the subcollection issue is an easy fix, but I'd like to address that in a separate PR, too, since it's technically tangential to this PR. Can we open a broad issue for child-objects?

hortongn:
@jamesvanmil Nice work!

PR #163, opened 2018-04-06
Add virus scan to js uploader

### LINKED ISSUE #56: Scan uploaded files for malware before ingest
### Descriptive summary

When a user uploads a file and saves the work, all files should be scanned for viruses.  If a virus is found, it should not be attached to the work and the user should be notified.

This feature can be implemented the way as it's done on Scholar@UC:
https://github.com/uclibs/scholar_uc/commit/8161ecfdb278b91c93390ca74c38680f288f3496

Trello: https://trello.com/c/QFKaKwJD/6-virus-screening-of-deposited-files

crowesn:
Fixes #56 

Overrides Hyrax gem to add virus scan validation to uploaded_file model as well as exception handling to controller. JS uploader now displays contextual error message when a virus is detected at time of upload. Disables scan in the file attachment job so that we don't scan twice.

* Adds virus scan validation
* Add exception handling and UI error message for infected files
* Disable virus scan in file attach job
* Adds virus scan specs


app/controllers/hyrax/uploads_controller.rb
```diff
@@ -0,0 +1,20 @@
+# Copied from hyrax 366d146
+
+module Hyrax
+  class UploadsController < ApplicationController
```
hortongn: @crowesn Since you are only changing the #create method, can you reopen the UploadsController class instead of completely overriding it?  That will help keep us from having to update this file for future versions of Hyrax.  Reopening the class will also help with Coveralls.  Right now Coveralls sees the #destroy method as untested code.

app/controllers/hyrax/uploads_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    load_and_authorize_resource class: Hyrax::UploadedFile
+
+    def create
+      @upload.attributes = { file: params[:files].first,
+                             user: current_user }
+      @upload.save!
+    rescue ActiveRecord::RecordInvalid
+      render json: { files: [{ name: @upload.file.filename, size: @upload.file.size, error: @upload.errors[:base].first }] }
```
hortongn: @crowesn The test suite is never hitting your `render json` line.  We need a test for that.  If you can't get your feature spec to hit that line, we probably need a controller spec for #create that does.

app/models/hyrax/uploaded_file.rb
```diff
@@ -0,0 +1,27 @@
+# Copied from hyrax 366d146
+
+module Hyrax
+  # Store a file uploaded by a user. Eventually these files get
+  # attached to FileSets and pushed into Fedora.
+
+  class CheckVirusValidator < ActiveModel::Validator
+    def validate(record)
+      return true unless Hydra::Works::VirusCheckerService.file_has_virus?(record.file)
+      record.errors[:base] << "Virus detected in file "
```
hortongn: @crowesn The test suite is never hitting your `record.errors` line.  If you can't get your feature spec to hit that line, you should create a model spec that hits it.

hortongn:
@crowesn I made some comments on your PR inline with the code.  The virus checking worked really well for me.  I didn't find a situation where it behaved unexpectedly.  So see my other comments about reopening UploadsController and adding more test coverage.

app/controllers/hyrax/uploads_controller.rb
crowesn: Would I put this in an initializer?

hortongn: Put `
require Hyrax::Engine.root.join('app/controllers/hyrax/uploads_controller.rb')` as the first line of file.  Then all you need to include inside the class is the one method you are changing.

See https://github.com/uclibs/ucrate/blob/develop/app/controllers/hyrax/dashboard/profiles_controller.rb as an example where we reopened the ProfilesController class to override just the user_params method.

PR #164, opened 2018-04-10
[WIP] change seeds file

### LINKED ISSUE #151: The db:seed rake task needs fixin
### Descriptive summary

I noticed a couple of problems with the recently added `rake db:seed` task:

1) The works that are generated are not assigned an admin set.  This leads to problems when the owner tries to do something with the work.  All works should be assigned to the default admin set.

2) The generated admin user didn't have admin privileges when I tried it.  So something is broken there.

changy66:
Fixes #151

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -39,7 +39,7 @@ class AddSeedObjects < ActiveRecord::Migration[5.1]
     last_name: 'User',
     password: password,
     password_confirmation: password)
-  admin = Role.create(name: 'admin')
+  admin = Role.find_or_create_by(name: 'admin')
   admin.users << admin_user
   admin.save
   puts "Account created: #{admin_user.email}\n\n"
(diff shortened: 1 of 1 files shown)


hortongn:
I worked with Olivia on this PR.  So someone else should probably review.

crowesn:
@hortongn Should we need to pre-load the default admin set before running?

crowesn:
@hortongn @changy66 

Issue I'm encountering -
`rake db:seed` runs as expected, creates 3 users and works for all users. However, the catalog is empty after seed task is run and when you log in as one of the users, works are visible via dashboard but users doesn't have edit rights.

Going into the console and saving the work resolves the issue.

I also tried creating a work manually in the console, running this [block](https://github.com/uclibs/ucrate/pull/164/files#diff-f7a25bc30a409fc77050703fc048ca8cR52) and found that the work won't display in catalog unless saved twice.

hortongn:
@crowesn I'm seeing the same behavior as you after running the seeds, but I haven't had a chance to dig in and figure out why yet.

PR #198, opened 2018-05-04
Show profile fields on the user show page

### LINKED ISSUE #188: Show profile fields on the user show page
### Descriptive summary

The work in https://github.com/uclibs/ucrate/pull/173 adds new fields to the user **edit** profile page, but those fields don't show on the user **show** page.

Add fields on the user show page so they display in the lower-left column.  They should display the same way they do on Scholar profile pages.  Fields should only display if they have content in them.

See the example at
https://scholar.uc.edu/users/hortongn@ucmail-dot-uc-dot-edu

Depends on https://github.com/uclibs/ucrate/pull/173 being merged first.

changy66:
Fixes #188 

Pulled the file 'user_info.html.erb' from hyrax
Changed 'department' to 'ucdepartment' so the department fields can show on the user show page.

### DIFF
--- a/app/views/hyrax/users/_user_info.html.erb
+++ b/app/views/hyrax/users/_user_info.html.erb
@@ -1,0 +1,81 @@
+<dl id="user_info">
+
+<% if user.orcid.present? %>
+  <dt><%= orcid_label('profile') %></dt>
+  <dd><%= link_to user.orcid, user.orcid, { target: '_blank' } %></dd>
+<% end %>
+
+<% if Hyrax.config.arkivo_api? && user.zotero_userid.present? %>
+  <dt><%= zotero_label(html_class: 'profile') %></dt>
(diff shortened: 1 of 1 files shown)


app/views/hyrax/users/_user_info.html.erb
```diff
@@ -0,0 +1,81 @@
...
+    <dt>Office</dt>
+    <dd><%= user.office %></dd>
+  <% end %>
+
+  <% if user.address %>
+    <dt><i class="glyphicon glyphicon-map-marker"></i> Address</dt>
+    <dd><%= user.address %></dd>
+  <% end %>
+
+  <% if user.affiliation %>
```
hortongn: @changy66 we should be using `user.uc_affiliation` instead of `user.affiliation` (on both lines 71 and 73)

PR #214, opened 2018-05-15
Fix flaky tests that fail unreliably

### LINKED ISSUE #213: Fix flaky tests
### Descriptive summary

Tests are failing after merging PR #178 

1. Tests are failing for seeds 42315, 55877 etc...
1. bundle exec rspec  --seed 42315

### Related work

#148

jayanthdeejay:
Fixes #213

This PR aims to fix some of the flaky tests that fail when run in specific order. Introduced [here](https://github.com/uclibs/ucrate/blob/b31fb43fb113ac8ed4381f09177f8c8e09237824/spec/features/welcome_mailer_spec.rb#L11).

Tests fail when run in the following order:
```ruby
rspec './spec/features/uc_shibboleth_spec.rb[1:3:1,1:4:2]' './spec/features/welcome_mailer_spec.rb[1:3:1]' --seed 42315
```

* Revert any changes made to test configurations in the same examples where they were changed in the first place.

### DIFF
--- a/spec/features/uc_shibboleth_spec.rb
+++ b/spec/features/uc_shibboleth_spec.rb
@@ -70,7 +70,6 @@ describe 'UC account workflow', type: :feature do
 
   describe 'overridden devise sign-in page' do
     it 'shows a shibboleth login link if shibboleth is enabled' do
-      AUTH_CONFIG['shibboleth_enabled'] = true
       visit new_user_session_path
       if yaml['test']['shibboleth_enabled'] == true
         expect(page).to have_link('Central Login form', href: '/users/auth/shibboleth')
@@ -110,6 +109,7 @@ describe 'UC account workflow', type: :feature do
 
       it 'shows the local log in page' do
         page.should have_field('user[email]')
+        AUTH_CONFIG['shibboleth_enabled'] = true
       end
     end
   end
--- a/spec/features/welcome_mailer_spec.rb
+++ b/spec/features/welcome_mailer_spec.rb
@@ -19,6 +19,7 @@ describe WelcomeMailer do
   after do
     ActionMailer::Base.deliveries = []
     User.find_by_email(user_email).delete
+    AUTH_CONFIG['signups_enabled'] = false
   end
   it 'sends welcome email to user upon registration' do
     email.to.should eq([user_email])


spec/features/uc_shibboleth_spec.rb
```diff
@@ -70,7 +70,6 @@
 
   describe 'overridden devise sign-in page' do
     it 'shows a shibboleth login link if shibboleth is enabled' do
-      AUTH_CONFIG['shibboleth_enabled'] = true
```
hortongn: @jayanthdeejay I think we need to keep `AUTH_CONFIG['shibboleth_enabled'] = true` here.  This test requires `shibboleth_enabled` to be `true` for the test to work properly.  

I think you can add it back in and the specs will pass even when using seed `42315`

jayanthdeejay: @hortongn I agree. I removed this line because, by [default](https://github.com/uclibs/ucrate/blob/559ecbc4b0b68f1c52c1cd5fd4939bd7b494bce1/config/authentication.yml#L6), `shibboleth_enabled` is true. Only for [this](https://github.com/uclibs/ucrate/blob/0cef54ea4aee3a272ee980a34e38a221dd721123/spec/features/uc_shibboleth_spec.rb#L104-L115) example, `shibboleth_enabled` is set to false and I re-enabled it in the same example so that any diversion from test environment configuration are restored for other examples to run with default test configurations. 

Also did the same for [this](https://github.com/uclibs/ucrate/blob/0cef54ea4aee3a272ee980a34e38a221dd721123/spec/features/welcome_mailer_spec.rb#L5) example. [Diversion](https://github.com/uclibs/ucrate/blob/0cef54ea4aee3a272ee980a34e38a221dd721123/spec/features/welcome_mailer_spec.rb#L11) from default is [restored](https://github.com/uclibs/ucrate/blob/0cef54ea4aee3a272ee980a34e38a221dd721123/spec/features/welcome_mailer_spec.rb#L22). 

I think, we can even remove [this](https://github.com/uclibs/ucrate/blob/0cef54ea4aee3a272ee980a34e38a221dd721123/spec/features/uc_shibboleth_spec.rb#L94) line.

PR #254, opened 2018-06-26
Port over Scholar's ETD manager functionality

### LINKED ISSUE #248: Port over Scholar's ETD manager functionality
### Descriptive summary

Scholar only shows the "ETD" work type to admin users and users with the "ETD Manager" role.  We want to replicate this feature in UCrate, but using the new `hydra-role-management` method that we currently use for admin users.  So we want ETD managers to be added through the database instead of using a yaml file.

Done looks like:
* When creating a new work, the ETD work type is only offered if the current user has the ETD Manager role.
* The current user can only change existing ETD works if the user has the ETD Manager role.

Resources:
* How we make an admin user in UCrate: https://github.com/uclibs/ucrate/wiki/Setting-up-a-development-app-for-an-admin-user
* Scholar commit that added ETD manager: https://github.com/uclibs/scholar_uc/commit/7a02ba6367c9d49f65363f24ae2141ddf904b928

changy66:
Fixes #248

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -132,3 +132,11 @@ Style/SingleLineBlockParams:
 
 Style/SymbolArray:
   Enabled: false
+
+Metrics/AbcSize:
+  Exclude:
+    - 'app/models/ability.rb'
+
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - 'app/models/ability.rb'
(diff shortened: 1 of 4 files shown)


hortongn:
@changy66 You'll need to rebase this to get the solr_wrapper fix we added.

PR #256, opened 2018-06-27
Reopened visibility file from the engine, updated default to public.

### LINKED ISSUE #250: New works should default to public visibility
### Descriptive summary

When creating a new work, the visibility should default to `Public`

You can see how this was implemented in Scholar here:
https://github.com/uclibs/scholar_uc/commit/1e5d650679ff315c977d97444d758b60f94d7e01

(you might want to just re-open https://github.com/samvera/hydra-head/blob/master/hydra-access-controls/app/models/concerns/hydra/access_controls/visibility.rb#L20-L28 instead of using an initializer)

Be sure to include a test.

chuckgreenman:
Refactor using the re-opening technique, instead of using whole file.

Added tests for create work and batch create works.

Comply with rubocop.

Provide explicit conversion of path to string.

Path figured out, tests working?

Add method guard, fix test for batch create spec.

Fixes #250 ;

Changed the visibility method so that it defaults to open rather than private.

``` ruby
# frozen_string_literal: true

require "pathname"
visibility_path = Pathname.new(Gem.loaded_specs['hydra-access-controls'].full_gem_path + '/app/models/concerns/hydra/access_controls/visibility.rb')
require visibility_path
module Hydra::AccessControls
  module Visibility
    def visibility
      return AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC if id.nil?
      if read_groups.include? AccessRight::PERMISSION_TEXT_VALUE_PUBLIC
        AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
      elsif read_groups.include? AccessRight::PERMISSION_TEXT_VALUE_AUTHENTICATED
        AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED
      else
        AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE
      end
    end
  end
end
```

* Visibility should be open by default, instead of private.
* Added tests

### DIFF
--- a/app/models/concerns/hydra/access_controls/visibility.rb
+++ b/app/models/concerns/hydra/access_controls/visibility.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "pathname"
+visibility_path = Pathname.new(Gem.loaded_specs['hydra-access-controls'].full_gem_path + '/app/models/concerns/hydra/access_controls/visibility.rb')
+require visibility_path
+module Hydra::AccessControls
+  module Visibility
+    def visibility
+      return AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC if id.nil?
+      if read_groups.include? AccessRight::PERMISSION_TEXT_VALUE_PUBLIC
+        AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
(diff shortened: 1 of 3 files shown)


chuckgreenman:
Travis failing, solr_wrapper not yet updated for change in how md5 is handled.

PR #322, opened 2018-07-26
[WIP] Configures facets to display the same as in Scholar

### LINKED ISSUE #245: Configure facets to display the same as in Scholar
### Descriptive summary

When on the /catalog page, the facets displayed on the left side of the page are different or labeled differently than on Scholar.  Configure the facet display to match Scholar as closely as possible.

We do, however, want to keep the new "Collection" facet at the bottom of the facets list.

Most of the changes will happen here:
https://github.com/uclibs/ucrate/blob/develop/app/controllers/catalog_controller.rb#L42-L54

chuckgreenman:
Fixes #245 ;

Configures facets to display the same as in Scholar

Tests explicitly expect the facets we want on the page, they also expect exactly 6 facets, instead of explicitly expecting a page not to have facets.  This gets around a scenario where the facet name we expect not to be there is present, but not the same display name due to translation or something in that vein.

* Configures facets to display the same as in Scholar

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,6 +71,7 @@ RSpec/ExampleLength:
     - 'spec/features/browse_dashboard_works_spec.rb'
     - 'spec/features/browse_catalog_spec.rb'
     - 'spec/features/catalog_search_spec.rb'
+    - 'spec/features/catalog_facet_spec.rb'
     - 'spec/features/collection_export_spec.rb'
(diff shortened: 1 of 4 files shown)


scherztc:
- [ ] Type of Work
- [ ] Creator / Author
- [ ] Subject
- [ ] College
- [ ] Language
- [ ] Publisher
- [ ] Date Created

Issue #310, opened 2018-07-24
Resolve Coveralls flakiness and coverage issues


hortongn:
### Descriptive summary

Figure out why Coveralls drops coverage on some PRs.  Also resolve coverage issues for `/lib/tasks/embargo_manager.rake`

jamesvanmil:
Closed with #340 and #342

PR #325, opened 2018-07-28
Add rspec-retry gem and fix batch_edit_spec

hortongn:
Partial fix for #310 

`spec/features/batch_edit_spec.rb:85` has been failing pretty frequently.  Samvera folks added a `wait: 5` to the test, but it doesn't seem to be enough.  Sometimes the batch takes a **long** time to save and the test gives up.  Increasing the wait to 1 minute helps some, but we don't want to always wait that long.

So I added the `rspec-retry` gem that lets us configure certain test to automatically run again and give them another chance to pass.  I've configured `spec/features/batch_edit_spec.rb:85` to try running 3 times before giving up.  After running it locally and on Travis many times, that spec has always passed.  So this should get rid of false failures for that test.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,7 @@ group :test do
   gem 'database_cleaner'
   gem 'factory_bot_rails'
   gem 'rspec-activemodel-mocks'
+  gem 'rspec-retry'
   gem 'selenium-webdriver', '3.12.0'
   gem 'shoulda-matchers', '~> 3.1'
 end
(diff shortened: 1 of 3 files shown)


hortongn:
This is now ready for review.

chuckgreenman:
@hortongn Looks good and works well locally.  Capybara does have a wait_until { something.visible? } method though, is it worth exploring that to reduce our reliance on retrying?  If not I'll just approve and merge.

hortongn:
`wait_until` was actually removed in capybara 2.0.  So it's not an option anymore ☹️

PR #334, opened 2018-08-02
Redirect dashboard profile to user profile

### LINKED ISSUE #309: Stop using the "dashboard" version of the user profile page
### Descriptive summary

Hyrax has a "dashboard" version of the user profile show page AND a "normal" version of the same page.

For an example see
http://curly.libraries.uc.edu:3002/dashboard/profiles/admin@example-dot-com?locale=en
and
http://curly.libraries.uc.edu:3002/users/admin@example-dot-com?locale=en

The dashboard version can be accessed via Dashboard -> Your activity -> Profile
The normal version can be access via Home -> People -> (select the user)

To avoid confusion, we want to stop using the "dashboard" version of the page.  Any places that link to `/dashboard/profiles/...` should instead link to `/users/...`

crowesn:
Fixes #309 

Add redirect to user profile in dashboard profile show action.

* Tweak dashboard profile controller
* Add spec


spec/controllers/hyrax/dashboard/profiles_controller_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+    expect(response).to redirect_to "/users/#{user.user_key.gsub('.', '-dot-')}?locale=en"
+  end
+
+  it "redirects to root if user does not exist" do
+    expect do
+      get :show, params: { id: 'johndoe666' }
+    end.to raise_error ActiveRecord::RecordNotFound
+  end
+
+  describe "#edit" do
```
hortongn: This works great and I think a redirect in the controller is the most elegant solution for us.  Only change I'm requesting is in the spec file.  Can you remove the `describe "#edit"` and `describe "#update"` blocks?  We didn't change those actions and don't need the tests in our app.  Might be able to remove those fixtures too.

crowesn:
@hortongn this should be ready to go.

Issue #317, opened 2018-07-26
Port over search fields, facet fields, display fields, and show fields from Scholar 3


scherztc:
### Descriptive summary

Compare the app/controllers/catalog_controller.rb files in Scholar and UCrate

### Expected behavior

Keyword Search has all descriptive metadata fields.
Facets should be ported over from Scholar 3. - https://github.com/uclibs/ucrate/issues/245
Display Fields should display as Scholar 3 and link as Scholar 3.
Show Fields should display as Scholar 3 and link as Scholar 3

### Actual behavior

Default Hyrax Behavior.

This isssue will be related to multiple other ux issues.

scherztc:
https://github.com/uclibs/ucrate/issues/245

newmanld:
Changes that I recommend that will be different in Scholar 4 than Scholar 3.  (Based on earlier discussions with P.O.s and/or Dev team although feedback has not always been clear or complete.)

**Facets** - restore Date Created facet  (Used to exist as a facet, we think it was accidentally removed.  May have something to do with the 'alt_date_created' confusion.)

**Catalog Browse** - reduce the fields that are listed, from any work type including Collections, to these fields, when they exist (have data) on a work:  
Title    (maintain offset placement)
Type  (same as Type of Work facet)
Description/Abstract
Creator/Author(s)
Submitter 
Date Uploaded
Date Modified
Date Created
Rights 

(Many other fields currently display on the Catalog Browse in Scholar 3, making it too lengthy.  Not all fields apply to all work types - Collection has the fewest and may not have Date Uploaded or Date Modified.)
 
**Catalog Browse Sort:** UCRATE does not include Title (A-Z and Z-A).  If not already in a separate issue (one of our local overrides?), can adding Title sort to Catalog Browse be included in this GitHub Issue?

**Work Show page:**  Add Date Uploaded and Date Modified.  I don't have a specific suggestion on placement - suggest sending the Dev team a screenshot of a mockup.

newmanld:
Per discussion with @chuckgreenman - the field labeled 'Rights' in Scholar 3 /hyrax 1.x is now labeled 'License' in Scholar 4 / hyrax 2.x.   A new field with different values is labeled 'Rights' in Scholar 4 / hyrax 2.x.  I recommend keeping the label of 'License' for the existing data, keeping that field in the Catalog Browse, and not adding the new 'Rights' field to Catalog Browse.

hortongn:
[...]

PR #337, opened 2018-08-06
Bug/#317 new catalog controller approach

chuckgreenman:
refs #317 

This pull request implements changes to the catalog controller so that the search page displays the correct fields.  As it stands, the code works in our development environment, however, fields involving dates are not displayed correctly in the testing environment.

We can make fields showing dates appear correctly in the test environment, but we have to remove the `type: :date` flag in solr_name.  That functionality can be had with this line:

config.add_index_field solr_name("date_modified", :stored_sortable), helper_method: :human_readable_date

Tests for the date fields exist, but they are prefixed with skip for now.

* Add, remove, and reorder fields on the search result page for items to make page similar to scholar.


chuckgreenman:
Coverage down a bit, doesn't seem connected to changes that we're made.  Rebased, resolved conflicts.

chuckgreenman:
Changed line, tried saving with just a yyyy and with mm-dd-yyyy format.  It works!

chuckgreenman:
@hortongn Restarting the failed travis build resolved the problem, I think this is good to go!

PR #355, opened 2018-08-16
Modifies batch upload to include the correct metadata.

### LINKED ISSUE #304: Modify batch upload to behave the same as in Scholar
### Descriptive summary

Make the needed changes to have the batch upload experience in UCrate (Dashboard -> Works -> Create batch of works) be the same as in Scholar.

Compare the steps and screens with Scholar.  Also make sure that the works generated by the batch upload are consistent with Scholar.

scherztc:
Fixes #304 

Overwrites hyrax batch import to get the correct metadata fields for all work types on batch upload.

This will get batch upload to work as it is in Scholar.


scherztc:
@hortongn Please review this again.  Since publisher is a default field in hyrax we removed etd_publisher.

PR #378, opened 2018-09-07
Modifies batch edit to set a base values.

### LINKED ISSUE #305: Modify batch edit to behave the same as in Scholar
### Descriptive summary

Make the needed changes to have the batch edit experience in UCrate be the same as in Scholar.

Compare the steps and screens with Scholar. We should be able to edit the exact same metadata fields in UCrate as we do in Scholar.  Also make sure that the works changed by the batch edit are consistent with Scholar.

https://scholar-dev.uc.edu is a good place to test batch edit for Scholar.

scherztc:
Fixes #305 

Batch edit works with the following fields.  We will still need to create different issues for some of the other fields.


hortongn:
@scherztc Please fix the merge conflict

scherztc:
@hortongn .  Fixed the merge conflict, but coveralls decreased.

Issue #366, opened 2018-08-30
Update values in the work license dropdown


hortongn:
### Descriptive summary

The dropdown menu that displays on the `License` field for new works has values that don't match the corresponding field in production Scholar (in Scholar it is named the "Rights" field).

The correct values in production Scholar are
```
<option value="http://rightsstatements.org/vocab/InC/1.0/">All rights reserved</option>
<option value="http://creativecommons.org/licenses/by/4.0/">Attribution 4.0 International</option>
<option value="http://creativecommons.org/licenses/by-sa/4.0/">Attribution-ShareAlike 4.0 International</option>
<option value="http://creativecommons.org/licenses/by-nc/4.0/">Attribution-NonCommercial 4.0 International</option>
<option value="http://creativecommons.org/licenses/by-nd/4.0/">Attribution-NoDerivs 4.0 International</option>
<option value="http://creativecommons.org/licenses/by-nc-nd/4.0/">Attribution-NonCommercial-NoDerivs 4.0 International</option>
<option value="http://creativecommons.org/licenses/by-nc-sa/4.0/">Attribution-NonCommercial-ShareAlike 4.0 International</option>
<option value="http://creativecommons.org/publicdomain/mark/1.0/">Public Domain Mark 1.0</option>
<option value="http://creativecommons.org/publicdomain/zero/1.0/">CC0 1.0 Universal</option>
<option value="http://www.opendatacommons.org/licenses/pddl/1.0/">Open Data Commons Public Domain Dedication and License (PDDL)</option>
<option value="http://www.opendatacommons.org/licenses/by/1.0/">Open Data Commons Attribution License (ODC-By)</option>
<option value="http://opendatacommons.org/licenses/dbcl/1.0/">Open Data Commons Open Database License (ODbL)</option></select></div>
```

Update UCrate so the values are the same as above.

PR #386, opened 2018-09-11
Fix missing license wizard button

### LINKED ISSUE #365: License wizard button is missing
### Descriptive summary

The license wizard was ported over from uclibs/scholar_uc in https://github.com/uclibs/ucrate/pull/316.  However, it has gone missing from the new/edit work pages.  So we need to figure out happened and get the button back (and working)

crowesn:
Fixes #365 
Depends on #366 

The JS for the license wizard was missing a div in the view to render the button. I added a div class and simplified the style. Also added some feature specs to check that the button is showing where it should.

### DIFF
--- a/app/assets/javascripts/select-license.js
+++ b/app/assets/javascripts/select-license.js
@@ -13,7 +13,7 @@ var ls = function() {
             name: 'All rights reserved',
             priority: 1,
             available: true,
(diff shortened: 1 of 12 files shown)


crowesn:
@hortongn again not so sure about coveralls at this one - I added feature tests and I don't think I introduced any new classes or modules.

PR #395, opened 2018-09-12
Fileset edit rights

### LINKED ISSUE #362: Work transfer does not reset edit rights on attached file_sets
### Descriptive summary

Work transfer does not reset edit rights on attached file_sets when the receiving user opts to remove depositor access.

### Expected behavior

When transferring a work, the receiving user can choose to remove access original depositor. When they do this the original depositor should not be able to edit the work or the attached file_sets.

### Actual behavior

When the receiving user chooses to remove access from the original depositor, the original depositor is still listed as an edit user of attached file_sets.

1. Create a work with a file
1. Transfer it to another user
1. As the other user, accept the request, removing depositor access
1. Check the edit_users on the attached file_sets

### Related work

See this class:

https://github.com/samvera/hyrax/blob/master/app/services/hyrax/change_content_depositor_service.rb

Should this be in the `work.file_sets.each` block?:

```ruby
f.apply_depositor_metadata(user)
f.permissions = [] if reset
f.save!
```

Note that this is an issue in Hyrax - we should open an issue there an contribute a fix.

crowesn:
Fixes #362 

Work transfer does not reset edit rights on attached file_sets when the receiving user opts to remove depositor access.

Override change_content_depositor_service to clear the permissions for attached filesets as well.

```ruby
      work.file_sets.each do |f|
        f.permissions = [] if reset
        f.apply_depositor_metadata(user)
        f.save!
      end
```
* Add override `app/services/hyrax/change_content_depositor_service.rb`
* Add override `spec/services/hyrax/change_content_depositor_service_spec.rb`
* Add `f.permissions = [] if reset` to fileset block and tweak specs.

### DIFF
--- a/app/services/hyrax/change_content_depositor_service.rb
+++ b/app/services/hyrax/change_content_depositor_service.rb
@@ -1,0 +1,18 @@
+module Hyrax
+  class ChangeContentDepositorService
+    # @param [ActiveFedora::Base] work
+    # @param [User] user
+    # @param [TrueClass, FalseClass] reset
+    def self.call(work, user, reset)
+      work.proxy_depositor = work.depositor
+      work.permissions = [] if reset
+      work.apply_depositor_metadata(user)
+      work.file_sets.each do |f|
+        f.apply_depositor_metadata(user)
+        f.save!
+      end
+      work.save!
+      work
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


crowesn:
I think the coveralls score on here is bunk. Come at me.

PR #410, opened 2018-09-17
Check for presence and whitelist payload_concern param on batch upload

### LINKED ISSUE #406: Verify payload_concern parameter on the batch upload form
### Descriptive summary

When creating a new batch upload, the user has to pick a work type and then lands on the batch create form with a `payload_concern` param that indicates the type of works to create.

Example URL:
`/batch_uploads/new?locale=en&payload_concern=Article`

However, if someone accesses the form without the payload_concern (i.e. just `/batch_uploads/new?locale=en` it generates an error because the form tries to use the payload_concern value.

We probably want to add some code to `Hyrax::BatchUploadsController#new` that 1) verifies the `payload_concern` param is present and 2) verifies the value is a legitimate work type.  If this fails, we should just return a 404.

crowesn:
Fixes #406 

Checks for presence and validity of payload_concern params on batch upload :new action. If the param is absent or not valid, show a `Not Found` routing error.

I attempted to override the new action in the controller but had difficulty because this is very form oriented and any new action override would result in error.

* Add form builder function override to batch_uploads_controller.rb
* Add test to batch_uploads_controller_spec.rb

### DIFF
--- a/app/controllers/hyrax/batch_uploads_controller.rb
+++ b/app/controllers/hyrax/batch_uploads_controller.rb
@@ -18,5 +18,16 @@ module Hyrax
                                      attributes_for_actor.to_h.merge!(model: klass),
                                      operation)
       end
+
+      # Hyrax override to check for presense and validity of payload_concern param
+      def build_form
+        super
+        raise ActionController::RoutingError, 'Not Found' unless work_type_specified?
+        @form.payload_concern = params[:payload_concern]
+      end
+
+      def work_type_specified?
+        Hyrax.config.registered_curation_concern_types.include? params[:payload_concern]
+      end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/hyrax/batch_uploads_controller.rb
```diff
@@ -18,5 +18,12 @@ def create_update_job(klass)
                                      attributes_for_actor.to_h.merge!(model: klass),
                                      operation)
       end
+
+      # Hyrax override to check for presense and validity of payload_concern param
+      def build_form
+        super
+        raise ActionController::RoutingError, 'Not Found' unless Hyrax.config.registered_curation_concern_types.include? params[:payload_concern]
```
hortongn: This works great.  I have one small suggestion.  

Move `Hyrax.config.registered_curation_concern_types.include? params[:payload_concern]` into a new method called something like `work_type_is_specified?`.  

Then the raise line will read a little better with 
`raise ActionController::RoutingError, 'Not Found' unless work_type_is_specified?`

PR #421, opened 2018-09-18
Update to hyrax 2.2.1

### LINKED ISSUE #363: Research Hyrax 2.2.1 update
### Descriptive summary

We need to decide if we should update to Hyrax ~~2.2.0~~ 2.2.1.
https://github.com/samvera/hyrax/releases/tag/v2.2.1
https://github.com/samvera/hyrax/releases/tag/v2.2.0

First steps:
1. Update app to use Hyrax 2.2.1
1. Run the tests and note what fails
1. Discuss with team how much work it will be to upgrade

bsp3ars:
Fixes #363 

Updated Hyrax to 2.2.1. Updated the feature specs, the Gemfile, and a handful of other things.


bsp3ars:
Changed the views mentioned by @hortongn, and then came across a few issues that needed to be fixed. The missing css id issue from Hyrax 3038 still occasionally throws an error locally, but is not a cause for concern. Fixed the missing GID issue by adding a collection type 2 to catalog_index_spec.rb.

PR #439, opened 2018-09-26
Update Collection Form

### LINKED ISSUE #399: Update the collection form to match Scholar
### Descriptive summary

The new collection form needs changes so that the fields match up with the collection form in Scholar.

Scholar has the following fields on the collection form (all required)
* Title (single value)
* Creator (multiple value)
* Description (single value)
* Rights (which we are calling License in ucrate)

Those should be the only fields available on the form.  No `Additional fields` button.  (We may eventually want additional metadata fields on the collection form, but that's a future issue.)

Note: collection avatar and visibility are handled on other tabs and not part of this issue.

~~It also appears that the selected license is not being remembered when the collection is edited.  Editing a collection will display a form that looks like a license was never chosen.  That bug should be fixed as part if this issue.~~ See #438

bsp3ars:
Fixes #399 

Merged the collection form from Scholar to ucrate. Made all fields except creator single value (required pulling in the edit.html to fix the displayed title), moved the license wizard to the left side, and updated the specs.

### DIFF
--- a/app/forms/hyrax/forms/collection_form.rb
+++ b/app/forms/hyrax/forms/collection_form.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require Hyrax::Engine.root.join('app/forms/hyrax/forms/collection_form.rb')
+module Hyrax
+  module Forms
+    class CollectionForm
+      self.terms = [:title, :creator, :description, :license, :thumbnail_id, :representative_id, :visibility]
+
+      self.required_fields = [:title, :creator, :description, :license]
+
+      def self.model_attributes(_)
+        attrs = super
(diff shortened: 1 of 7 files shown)


app/forms/hyrax/forms/collection_form.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
```
hortongn: @bsp3ars Instead of overriding `app/forms/hyrax/forms/collection_form.rb`, I'd like for you to re-open the class instead.  It will reduce our technical debt for when we upgrade Hyrax in the future.  I can show you what I mean when you have a few minutes.

PR #444, opened 2018-09-28
render 404 error instead of template

### LINKED ISSUE #443: 404 template not rendering properly
### Descriptive summary

Object not found error raise in common objects controller is causing runtime error because we deleted the `public/404.html` file

https://github.com/uclibs/ucrate/blob/develop/app/controllers/common_objects_controller.rb#L13

crowesn:
Fixes #443 
Common objects controller is calling a template that was deleted in favor of custom error messages. Tweak so that it renders the proper message.


crowesn:
@hortongn This should fix the 404 travis failures for everything downstream.

PR #445, opened 2018-10-02
Fix collection license persistence

### LINKED ISSUE #438: Collection license is not saving
When creating a collection, the license that the user selects is not getting saved in the database, however it is getting passed (checked via params). After a further inspection of the problem and a discussion with @hortongn , it has become aware that the parameter :license is not getting allowed due to ```Unpermitted parameter: :license``` in the rails console. It most likely has something to do with strong parameters, however I didn't see any listed anywhere.

crowesn:
Fixes #438 

Collection license form values weren't persisting due to single/multi value issue. 
Add property override to local collection model and the problem is resolved.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -8,8 +8,10 @@ class Collection < ActiveFedora::Base
   include Hyrax::BasicMetadata
   self.indexer = Hyrax::CollectionWithBasicMetadataIndexer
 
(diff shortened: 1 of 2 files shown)


hortongn:
@crowesn This now has a merge conflict because PR #439 also touched `app/models/collection.rb`.  You can now remove your `self.multiple?` method since #439 already added it.

crowesn:
@hortongn this should be good to go.

PR #456, opened 2018-10-09
Delete the double quote if the description or note ends in double quote.

### LINKED ISSUE #441: A multi-line field that ends in a double-quote throws error
### Descriptive summary

A student had a problem saving a new work using content she had pasted from an Excel document into a Note field.  After further testing, it looks like multi-line fields on the new work form like Description and Note will throw an error if the contents contain 1) more than one line and 2) ends in a `"` char.

To reproduce, paste the following into the Description or Note field on a new work form:

```
test
test"
```

It will throw the following error when you try to save:
`Ldp::BadRequest in Hyrax::ArticlesController#create`
`RDF was not parsable: [line: 9, col: 1 ] Broken token (newline): ;`

Interestingly, the error seems to only occur for saving a new work.  Editing an existing work with the above content works fine.

We should try to catch this and escape it somehow before it tries to save.

changy66:
Fixes #441 

Create a method to replace '"' to '" ', if the string ends with a double quote. Use this method in the 'works_controller_behavior.rb' to make sure 'description' and 'note' fields don't end with a double quote.
Create a creating work spec file with the description and note ends with a double quote.


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -0,0 +1,28 @@
...
+          wants.html do
+            build_form
+            render 'new', status: :unprocessable_entity
+          end
+          wants.json { render_json_response(response_type: :unprocessable_entity, options: { errors: curation_concern.errors }) }
+        end
+      end
+    end
+
+    def prevent_fields_from_ending_in_double_quote
```
hortongn: @changy66 Let's talk about refactoring this method just a little bit.

hortongn:
@changy66 This is missing some test coverage of these lines:
https://github.com/uclibs/ucrate/blob/bug/%23%23441-A-multi-line-field/app/controllers/concerns/hyrax/works_controller_behavior.rb#L14-L19

We can talk about how to cover those lines with a test.

changy66:
Okay. Thanks.
On Oct 22, 2018, at 12:31 PM, Glen Horton <notifications@github.com<mailto:notifications@github.com>> wrote:


@changy66<https://github.com/changy66> This is missing some test coverage of these lines:
https://github.com/uclibs/ucrate/blob/bug/%23%23441-A-multi-line-field/app/controllers/concerns/hyrax/works_controller_behavior.rb#L14-L19

We can talk about how to cover those lines with a test.

PR #458, opened 2018-10-10
Allows for sorting and reverse title sorting alphabetically.

### LINKED ISSUE #333: Replicate Scholar's title sort on the catalog index page.
### Descriptive summary

On Scholar's /catalog page, the user can click `Sort by relevance` and sort results by title (ascending or descending).  We want to replicate this feature in ucrate.

Commit that implements this in Scholar: https://github.com/uclibs/scholar_uc/commit/75d268455f47a23193bdfe41931eff7ba5c94287

index catalog scholar uc

scherztc:
Fixes #333;

Replicates title sort on browse and search page.

Makes search and browse experience more like scholar.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,9 @@ Metrics/MethodLength:
     - 'app/controllers/classify_concerns_controller.rb'
     - 'app/services/works_report.rb'
     - 'app/forms/hyrax/forms/batch_upload_form.rb'
+    - 'spec/support/hyrax/factory_helpers.rb'
+    - 'app/indexers/hyrax/file_set_indexer.rb'
+    - 'spec/support/hyrax/helpers/factory_helpers.rb'
 
 Metrics/ClassLength:
   Exclude:
@@ -152,6 +155,11 @@ RSpec/VerifiedDoubles:
     - 'spec/views/hyrax/dashboard/collections/_show_actions.html.erb_spec.rb'
     - 'spec/helpers/hyrax_helper_spec.rb'
 
+RSpec/NamedSubject:
+  Exclude:
+    - 'spec/indexers/hyrax/collection_indexer_spec.rb'
+    - 'spec/indexers/hyrax/file_set_indexer_spec.rb'
+
 Style/ClassAndModuleChildren:
   Enabled: false
 
@@ -168,6 +176,10 @@ Style/SymbolArray:
 Metrics/AbcSize:
   Exclude:
     - 'app/models/ability.rb'
+    - 'app/indexers/hyrax/file_set_indexer.rb'
+    - 'app/indexers/hyrax/work_indexer.rb'
+    - 'app/models/ability.rb'
+    - 'app/indexers/hyrax/collection_indexer.rb'
 
 Metrics/CyclomaticComplexity:
   Exclude:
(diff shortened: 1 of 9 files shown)


app/indexers/hyrax/work_indexer.rb
```diff
@@ -0,0 +1,47 @@
...
+        admin_set_label = object.admin_set.to_s
+        solr_doc['admin_set_sim']   = admin_set_label
+        solr_doc['admin_set_tesim'] = admin_set_label
+        Solrizer.insert_field(solr_doc, 'sort_title', sortable_title(object.title.first), :stored_sortable) if object.title && !object.title.empty?
+      end
+    end
+
+    private
+
+      def sortable_title(title)
```
hortongn: I think we can dry this up a bit since both CollectionIndexer and WorkIndexer have the same exact `sortable_title` and `add_leading_zeros` methods.  We should be able to move those two private methods into the Hyrax::IndexesWorkflow module instead so they exist in just one place.  The re-open Hyrax::IndexesWorkflow and move the methods over.

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
```
hortongn: Do we actually need to override FileSetIndexer?  Seems like it wouldn't be needed to sort search results.

spec/indexers/hyrax/file_set_indexer_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
```
hortongn: Remove this spec if we don't need to be overriding FileSetIndexer.

.rubocop.yml
```diff
@@ -165,9 +173,18 @@ Style/SingleLineBlockParams:
 Style/SymbolArray:
   Enabled: false
 
+Style/GuardClause:
+  Exclude:
+    - 'app/indexers/hyrax/collection_indexer.rb'
```
hortongn: We shouldn't exclude this rubocop rule.  See my suggestion below for fixing this.

app/indexers/hyrax/collection_indexer.rb
```diff
@@ -0,0 +1,45 @@
...
+          (solr_doc['member_of_collection_ids_ssim'] ||= []) << col.id
+          (solr_doc['member_of_collections_ssim'] ||= []) << col.to_s
+        end
+      end
+    end
+
+    private
+
+      def sortable_title(title)
+        unless title.nil?
```
hortongn: Wrapping the entire method in an `unless` means that this method will return `nil` if we ever do have an empty title.  A better way to do it is 
```
def sortable_title(title)
  return '' if title.nil?
  cleaned_title = title.sub(/^\s+/i, "") # remove leading spaces
  cleaned_title.gsub!(/[^\w\s]/, "") # remove punctuation; preserve spaces and A-z 0-9
  cleaned_title.gsub!(/\s{2,}/, " ") # remove multiple spaces
  cleaned_title.sub!(/^(a |an |the )/i, "") # remove leading english articles
  cleaned_title.upcase! # upcase everything
  add_leading_zeros_to cleaned_title
end
```
that should get rid of the rubocop guard clause violation.

PR #461, opened 2018-10-11
Remove keyword from application spec test and locale files

### LINKED ISSUE #321: Remove keyword from application spec test and locale files.
### Descriptive summary

This field is inherited and available from Hyrax but we do not use it in our forms / presenters.  We use the subject field in a similar fashion.  There are still residual use of keyword in our spec test / locale and should be removed.

### Actual behavior

https://github.com/uclibs/ucrate/search?q=keyword&unscoped_q=keyword

1. Do this
1. Then do this...

### Related work

changy66:
Fixes #321 

Changed keyword to subject in spec files
Deleted keyword in locale files
Deleted keyword in collection presenter file


hortongn:
@scherztc Since you created the issue for this PR, I'm assigning it to you to review.  You'll have the best sense of whether the PR does what you intended.

PR #466, opened 2018-10-18
Implement autocomplete on specific new work form fields

### LINKED ISSUE #367: Research browser form autocomplete feature
### Descriptive summary

Some web browsers like Firefox will offer suggestions when filling out a form based on past content the user has entered into form fields.  It seems to be flakey in UCrate and often doesn't offer suggesting for new work fields.

Research if there are ways HTML forms can be configured to work better with autocomplete.  If we can do anything to make form fields in UCrate offer suggestions, we do it.

bsp3ars:
Fixes #367 

Modifies several partials in the app/views/records/edit_fields folder to have the attribute `autocomplete="on"` for several fields in create work forms.

The fields changed are:
* Advisor
* College
* Program or Department
* Publisher
* Subject
* Geo Subject
* Time Period
* Language

The Author and Creator fields are already pre-populated from the database, so they did not need to have autocomplete turned on (https://github.com/samvera/hyrax/wiki/Autocomplete-in-Hyrax)

Note: Since these fields are currently multi-value, I simplified the partials to only be multi-valued. If this changes in the future, you can change it by comparing it to the default partial in the hyrax-editor gem under `hydra-editor-4.0.2/app/views/records/edit_fields/_default.html.erb`.


bsp3ars:
Implemented changes to the create_etd spec that checks for the autocomplete fields

PR #469, opened 2018-10-19
Sets new collection visibility to public

### LINKED ISSUE #465: New collections should default to public visibility
### Descriptive summary

New collections in production scholar default to `Open Access` visibility when created.  We want the same behavior in ucrate (currently collections default to `Private` visibility).

Note: Hyrax 2.x doesn't even show the `Discovery` tab when creating a new work.  Users only see that tab after the collection has been created and then you go to the edit form.  So defaulting new collections to `Open Access` is not handled through the form.  The change may need to be made here:
https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/dashboard/collections_controller.rb#L112

bsp3ars:
Fixes #465

Sets new collections to Open Access (Public) by setting the value in the collection_form.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -118,6 +118,7 @@ RSpec/ExampleLength:
     - 'spec/forms/hyrax/batch_edit_form_spec.rb'
     - 'spec/jobs/batch_create_job_spec.rb'
     - 'spec/controllers/hyrax/batch_uploads_controller_spec.rb'
+    - 'spec/features/create_collection_spec.rb'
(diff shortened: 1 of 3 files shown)


hortongn:
@bsp3ars Works well, but it needs a test.  Something that just creates a collection and then verifies it has defaulted to open access after saving.

bsp3ars:
Added create_collection_spec which runs through the entire process of creating through the GUI. It checks for Open Access at the end.

PR #486, opened 2018-11-02
Fixes description field doesn't persist on generic work

### LINKED ISSUE #453: Description field doesn't persist on generic works
### Descriptive summary

When a generic work is created, the Description field is blank after the save.  It only happens with this work type and a recent commit likely introduced this bug.

1) Fix the description bug
2) Just in case there's another field with this problem somewhere, verify that all fields in all 8 work types properly save when works are created/edited.

changy66:
Fixes #453

### DIFF
--- a/app/forms/hyrax/generic_work_form.rb
+++ b/app/forms/hyrax/generic_work_form.rb
@@ -17,7 +17,7 @@ module Hyrax
     self.terms += %i[college department]
 
     ## Removing terms that we don't use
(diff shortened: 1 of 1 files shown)


hortongn:
@changy66 This PR has a merge conflict now.  Please rebase the next time you are in.  Thanks!

hortongn:
@changy66 This has a rubocop violation.

PR #542, opened 2018-11-29
Rename migrations to work with production scholar

### LINKED ISSUE #536: Fix migrations to work with scholar_uc database
### Descriptive summary

The existing migrations in db/migrate/ will not properly update the database on production Scholar when we deploy.  We need to tweak the migration files so the application understands which migrations have already been run and which have not.

hortongn:
Fixes #536 

Renames the migration files so that a `rake db:migrate` will properly update the production database.

I've run this against the scholar-qa database (Scholar 3.0) and everything migrated properly.

Let's wait until we've deployed to scholar-qa successfully to merge this.


hortongn:
Deployed this to scholar-qa and it successfully migrated the existing scholar_uc database.

PR #543, opened 2018-11-30
Use "Items in this Collection" on all collection pages

### LINKED ISSUE #528: Use "Items in this Collection" on collection show pages
### Descriptive summary

On collection show pages, production Scholar users the header "Items in this Collection" above the list of files.  We want to use that same text in ucrate.  Currently, ucrate uses the header "Works"

This might just require an override in hyrax.en.yml, but if not we should add it there so it can be translated.

bsp3ars:
Fixes #528

Adds Items in this Collection to the browse view portion of collections, not just the dashboard

### DIFF
--- a/spec/features/create_collection_spec.rb
+++ b/spec/features/create_collection_spec.rb
@@ -11,6 +11,15 @@ RSpec.describe 'Create a Collection', js: true do
       User.new(user_attributes) { |u| u.save(validate: false) }
     end
     let!(:user_collection_type) { create(:user_collection_type) }
+    let!(:user_collection) do
+      create(:public_collection,
+             user: user,
+             description: ['collection description'],
+             collection_type_settings: :nestable)
+    end
+    let!(:user_work) do
+      create(:work, title: ["King Louie"], member_of_collections: [user_collection], user: user)
+    end
 
     before do
       login_as user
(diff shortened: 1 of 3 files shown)


spec/features/create_collection_spec.rb
```diff
@@ -45,6 +45,7 @@
       expect(page).to have_link("Attribution-ShareAlike 4.0 International", href: 'http://creativecommons.org/licenses/by-sa/4.0/')
       expect(page).to have_content('Open Access')
       expect(page).to have_content('Items in this Collection')
+      expect(page).to have_content "Collection was successfully created."
```
hortongn: @bsp3ars This isn't the right string you want to test.  This spec file is checking the dashboard collections page.  We should probably also have it visit the public collection show page and verify the strings there as well.

spec/features/create_collection_spec.rb
```diff
@@ -45,6 +45,8 @@
       expect(page).to have_link("Attribution-ShareAlike 4.0 International", href: 'http://creativecommons.org/licenses/by-sa/4.0/')
       expect(page).to have_content('Open Access')
       expect(page).to have_content('Items in this Collection')
+
+      visit "/collections/#{id}"
```
hortongn: @bsp3ars The code works, but the spec is unfinished.  After `visit "/collections/#{id}"` you need to at least add another `expect(page).to have_content('Items in this Collection')` to make sure your change is showing on the public collection page.

PR #565, opened 2018-12-13
Adds support for Google Analytics.

### LINKED ISSUE #413: Implement statistics changes from Scholar
### Descriptive summary

We made several tweaks to scholar_uc to get statistics working for us.  So we need to...
1. Figure out what tweaks are still needed when using Hyrax 2.x
1. Implement those tweaks

The latest changes in scholar_uc to implement this can be seen here:
https://github.com/uclibs/scholar_uc/compare/1c59a0a3f57837308d1c755936691e0e52441594...0618cf55d046d869e73533ab4f70c8c306786bcd

A good first step is to bring over all those changes and get the specs passing.  We will then need to try testing it on scholar-qa.uc.edu

scherztc:
Fixes #413

This PR sets the variables needed for connecting to Google Analytics to provide view and download stats for each work and file_set. It also has the machinery for updating the totals in the dashboard.

The updating machinery includes rake task, database migrations, and copy scripts.

Refer to the notes on Legato for using in non-production environments :
https://github.com/samvera/hyrax/wiki/Analytics-workaround-for-non-production-environments

Refer to config_vault for authentication file and google API connection information:
https://git.uc.edu/UCLIBS/config_vault

### DIFF
--- a/app/models/hyrax/statistic.rb
+++ b/app/models/hyrax/statistic.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+module Hyrax
+  class Statistic < ActiveRecord::Base
+    self.abstract_class = true
+
+    class_attribute :cache_column, :event_type
+
+    class << self
+      include ActionDispatch::Routing::PolymorphicRoutes
+      include Rails.application.routes.url_helpers
+
+      def statistics_for(object)
+        where(filter(object))
+      end
+
+      def build_for(object, attrs)
+        new attrs.merge(filter(object))
+      end
+
+      def convert_date(date_time)
+        date_time.to_datetime.to_i * 1000
+      end
+
+      def statistics(object, start_date, user_id = nil)
+        combined_stats object, start_date, cache_column, event_type, user_id
+      end
+
+      # Hyrax::Download is sent to Hyrax::Analytics.profile as #hyrax__download
+      # see Legato::ProfileMethods.method_name_from_klass
+      def ga_statistics(start_date, object)
+        path = if object.class.to_s == 'FileSet'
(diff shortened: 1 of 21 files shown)


hortongn:
@scherztc This doesn't seem to be picking up file views.  See https://scholar-qa.uc.edu/users/hortongn@uc-dot-edu?locale=en

I remember you recently showing me something where the analytics code had the wrong URL for file set pages or something like that.  I wonder if we need to tweak something.  May be related to https://github.com/samvera/hyrax/issues/2718

hortongn:
@scherztc Also, I think you have some legitimate test failures because of a factory problem: https://travis-ci.org/uclibs/ucrate/builds/468055330?utm_source=github_status&utm_medium=notification

app/models/hyrax/statistic.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require Hyrax::Engine.root.join('app/models/hyrax/statistic.rb')
+
+module Hyrax
+  class Statistic < ActiveRecord::Base
+    def ga_statistics(start_date, object)
+      path = if object.class.to_s == 'FileSet'
+               hyrax_parent_file_set_path(object.parent, object)
```
scherztc: @hortongn :  I am testing that this override is actually working.  This should account for the different pathing on FileSets.

scherztc:
@hortongn :  I am also confused by not seeing any views here on the Analytics tab.  I am guessing that the tally is right, but something is going wrong with gathering the stats.

https://scholar-qa.uc.edu/files/cj82k728n/stats?locale=en

PR #571, opened 2018-12-20
Update Spanish and Mandarin translations

### LINKED ISSUE #390: Update the Spanish and Mandarin translations
### Descriptive summary

Note: Don't begin working on this issue until we are closer to release.  We will update all of the translations at once.

We should make sure that `config/locales/hyrax.es.yml` and `config/locales/hyrax.es.yml` contain the translated text for all of the entries in `config/locales/hyrax.en.yml`

Each work type also has it's own .es.yml and .zh.yml files that may need translations added.

Also browse through the application in Spanish and Mandarin mode and locate any other parts of the interface that are not translated.

hortongn:
Fixes #390

This fixes translations for the following pages
- homepage
- catalog index page
- work show pages
- collection show pages

You should be able to translate those pages and see that 99% of the text on them changes to another language.

The goal here was to do enough so that a non-logged-in user could use Scholar in another language.  Some more work is needed to translate other parts of the application (dashboard, edit pages, profiles, etc.)

It's not really practical for the reviewer to verify that all the translations are correct.  We will eventually want to have someone fluent in those languages try the application and see if there are any tweaks needed.


bsp3ars:
@hortongn 
You have a conflict with `config/locales/hyrax.en.yml`, as well as a misspelling in https://github.com/uclibs/ucrate/blob/b1268ec10ec7334a34e816b2614e431365da48ff/config/locales/hyrax.es.yml#L106 at the end for Scholar@UC

DOI's are also not translated, and I encountered a title error when I added a DOI to a existing work with Mandarin translations on
screen shot 2019-01-07 at 2 40 39 pm
screen shot 2019-01-07 at 2 37 22 pm

PR #576, opened 2019-01-10
Bug/#574 Some users are confused when a a new work says no files are attached

### LINKED ISSUE #574: Some users are confused when a a new work says no files are attached
### Descriptive summary

When using sidekiq for jobs, files are not immediately attached when a work is created.  The user instead see a page like the one below:

test

We've had some users (including experienced Scholar users) who are confused by the "This Document has no files associated with it." message at the bottom of the page.  They don't realize that their files will show up if they wait a minute and then refresh the page.

So we should change that message to say: 
"This <type> has no files associated with it.  (if you just created this <type>, it may take a few minutes for the files to display.  Try refreshing the page again.)  Otherwise, click "edit" to add more files."

You should be able to tweak it by adding https://github.com/samvera/hyrax/blob/master/config/locales/hyrax.en.yml#L416 to our own config/locales/hyrax.en.yml

Be sure to add a spec to verify the new text.

chuckgreenman:
Fixes #574;

Changed text of the message explaining files are being processed to avoid having users add multiple copies of the same files by accident.

### DIFF
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -59,6 +59,8 @@ en:
     base:
       form_progress:
         required_agreement: Check distribution license
+      items:
+        empty: 'This %{type} has no files associated with it. (If you just created this %{type}, it may take a few minutes for the files to display. Try refreshing the page again.) Otherwise, click edit to add more files.'
     batch_uploads:
       files:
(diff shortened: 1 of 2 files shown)


chuckgreenman:
Changed this so that the flash message includes the work type.  @hortongn

PR #580, opened 2019-01-14
Bug/#550 Resolves facet filtering bug

### LINKED ISSUE #550: Department facet should only show after College facet is selected
### Descriptive summary

Note: this bug is on production Scholar too and is therefore not a blocker for Scholar 4.0.

### Expected behavior

The Department facet should only show after the user has selected something from the *College* facet.

### Actual behavior

The Department facet shows after the user has selected something from *any* facet.

dept

chuckgreenman:
Fixes #550 ;

The department facet should not be shown unless the college facet has been set.  Previously, setting any facet would result in the department facet being shown.  This PR streamlines the filtered_facet_field_names method and adds a spec.

### DIFF
--- a/app/helpers/hyrax_helper.rb
+++ b/app/helpers/hyrax_helper.rb
@@ -85,16 +85,8 @@ module HyraxHelper
 
   def filtered_facet_field_names
     ## only show department if college is set in params
-    cache = facet_field_names
-    if params["f"].nil?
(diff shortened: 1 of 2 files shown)


hortongn:
@chuckgreenman You somehow ended up with 2 commits and 31 changed files in this PR.

PR #584, opened 2019-01-15
Fixes owner file permission problem with proxy deposit.

### LINKED ISSUE #488: Files created by proxies don't give edit access to owners
### Descriptive summary

If I'm a proxy and create a work with some files attached on behalf of another user (the owner), that owner will have edit access to the work, but not to the files.

The owner should be able to edit any file sets created by their proxies.

This is probably a bug in our app where we broke it somewhere, but we should verify that Hyrax doesn't have the bug.  You can test at https://nurax-dev.curationexperts.com

1. Create a work with file(s) as a proxy
1. Log in as the owner
1. Visit the show page for one of the files
1. Notice that you don't have an edit button

scherztc:
Fixes #488 and https://github.com/samvera/hyrax/issues/3416

Added a condition to the attach_files_to_work_job that adds the owner to the work_permissions map that gets used to set the file permissions.  This should be reviewed manually with config.active_job.queue_adapter = :sidekiq set in your environment.

We should also add a line in our manual test script to verify this also.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,6 +45,7 @@ Metrics/MethodLength:
     - 'spec/support/hyrax/helpers/factory_helpers.rb'
     - 'app/services/hyrax/actor_factory.rb'
     - 'app/services/hyrax/custom_stat_importer.rb'
+    - 'app/jobs/attach_files_to_work_job.rb'
 
 Metrics/ClassLength:
   Exclude:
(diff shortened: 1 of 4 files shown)


hortongn:
@scherztc One of your specs is trying to characterize files when Travis is building.  You probably need to add `expect(CharacterizeJob).to receive(:perform_later)` to that test so it skips characterization.  See your failed Travis builds for the specific test.

PR #625, opened 2019-01-28
Fix department facet

### LINKED ISSUE #609: Fix Department facet
### Descriptive summary

It looks like there's a bug with the recent PR (#550) to have the Department facet only display when the College facet is selected.

The Department facet appears to display all the time now.

bsp3ars:
Fixes #609

The code from #550 works, however after some testing I realized that the facet_field_names was not deleting when it was returning, so I looked over previous code and implemented an intermediary variable to return and make the changes to.


bsp3ars:
@hortongn I edited spec/features/catalog_facet_spec.rb and added 2 contexts to test the faceting

PR #631, opened 2019-01-29
add multiple? method to batch_upload_item.rb

### LINKED ISSUE #615: Batch create form should have scaler description and date_created fields
### Descriptive summary

When viewing the batch create form, the `description` and `date_created` fields allow users to enter multiple values for those fields.  However, both fields should only allow a single value (like on the regular create work forms).

We don't want the "Add another..." links to show for those fields.

new works by batch scholar uc

changy66:
Fixes #615 
add 'multiple?' method in batchUploadItem model and spec file for it.


scherztc:
@changy66 : It looks good for description, but we are still getting the error with date_created.

You attempted to set the property `date_created' of to a scalar value. However, this property is declared as being multivalued.

PR #632, opened 2019-01-29
Add last modified to work show pages

### LINKED ISSUE #600: Add a "last modified" date to work show pages
### Descriptive summary

Add a "Last modified" date to work show pages.  Scholar 3.x has this on the show page, but it is missing from 4.x.  

The format should be `Last modified: 11/08/2018`.  See the example screenshot below for where the text should appear.

In Scholar 3.x, app/views/hyrax/base/show.html.erb has
`<%= t('.last_modified', value: @presenter.date_modified) %>`

last_modified

bsp3ars:
Fixes #600

Adds the last modified section to the work show page

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -27,6 +27,7 @@
             <% end %>
             <%= render 'citations', presenter: @presenter %>
             <%= render 'social_media' %>
+            <%= t('.last_modified') %>: <%= @presenter.date_modified %>
           </div>
           <div class="col-sm-9">
             <%= render 'work_description', presenter: @presenter %>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/show.html.erb
```diff
@@ -27,6 +27,7 @@
             <% end %>
             <%= render 'citations', presenter: @presenter %>
             <%= render 'social_media' %>
+            <%= t('.last_modified') %>: <%= @presenter.date_modified %>
```
chuckgreenman: You're using a translation here, but last_modified isn't set in the locales YAML files.

bsp3ars: @chuckgreenman I believe that these translations come from hyrax, and will change based on our language selection. If not we can add them when we update the translations.

PR #644, opened 2019-02-06
Configures file attachment requirement for batch create.

### LINKED ISSUE #638: Creating a batch of works without files throws an error
### Descriptive summary

If you try to use batch create and only fill out the metadata (don't upload any files) Scholar will let you attempt to save the batch, but it will then throw an error:

```
NoMethodError in Hyrax::BatchUploadsController#create
undefined method `permit!' for nil:NilClass
```

I suspect that the `config.work_requires_files = false` in config/initializers/hyrax.rb is probably affecting both single works and batch works.  That is, we currently have Scholar configured to allow saving of batches without files.

We want users to be able to create single works without files, but at least one file should be required for batch works.

We probably need to tweak https://github.com/uclibs/ucrate/blob/develop/app/views/hyrax/base/_form_progress.html.erb#L11 so that it always requires files when batch create is being used.

This is probably a bug in Hyrax.  We should fix it in our app for the moment, but also check to see if there is an existing bug filed in Hyrax.  If not, open and issue and potentially submit the fix there as well.

scherztc:
Fixes #638 

The use case required that the batch create require a file to be attached while the work submission page without the file requirement.  To do this, I added a param check to see if we were coming from batch create.

### DIFF
--- a/app/views/hyrax/base/_form_progress.html.erb
+++ b/app/views/hyrax/base/_form_progress.html.erb
@@ -8,7 +8,7 @@
         <legend class="legend-save-work"><%= t('.requirements') %></legend>
         <ul class="requirements">
           <li class="incomplete" id="required-metadata"><%= t('.required_descriptions') %></li>
-          <% if Hyrax.config.work_requires_files? %>
+          <% if Hyrax.config.work_requires_files? || params[:controller] == "hyrax/batch_uploads" %>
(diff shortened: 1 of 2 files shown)


scherztc:
@hortongn :  I had to go with the spec/features/hyrax/batch_create_spec.rb test.  The view test you mentioned above is used by both single and batch create and it would have been too difficult to separate out the two scenarios in the one test.

PR #646, opened 2019-02-11
#634 Ensures that the locale will persist.

### LINKED ISSUE #634: Retain language choice when clicking all links
### Descriptive summary

Scholar remembers the user's language selection by appending `?locale=` to the end of URLs.

However, some of the links in our application don't include `?locale=` and the user's choice gets reset back to English.

Some of those links are:
* Clicking on the big "scholar@uc" link at the very top of the page
* Clicking on the link for an attached file on a work

Those links (and any others that drop the `?locale=`) should be tweaked to preserve the language choice.

chuckgreenman:
Fixes #634  ;

There are several places in the application where anchor tags are used rather than rails link_to helpers.  Helpers ensure that language selection persists.  This PR transitions relative links in the application to rails tags.

I found a missing translation in the course of working on this issue, I've added them to this PR as well.

### DIFF
--- a/app/views/_masthead.html.erb
+++ b/app/views/_masthead.html.erb
@@ -11,7 +11,9 @@
         </button>
         <%= render '/logo' %>
         <% unless controller_name == 'homepage' %>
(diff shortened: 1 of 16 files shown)


chuckgreenman:
@hortongn Cleaned up with rubocop, updated with changes we talked about.

PR #649, opened 2019-02-11
#608 Updates README and seeds file for admin role assignment

### LINKED ISSUE #608: Update README and seeds.rb admin role assignment method
Update the README and seeds.rb file to use a find or create method in the assignment of an admin

### Expected behavior
When running the seeds file, we want to create a user that is assigned the admin role.  When new developers are running through set up of a user with an admin role, we want to reduce confusion, as creation and assignment of a role that already exists fails somewhat quietly.

### Actual behavior
If the admin role already exists, the seed file will attempt to create the role and save the user to the role.  However, ActiveRecord returns an object of type role with a nil Id rather than failing out.

1. Create a role called admin
1. Create a new user
1. Run the seeds file
1. Attempt to log in as the admin user created with the password generated for you.
1. Navigate to the dashboard and look for admin options.

chuckgreenman:
Fixes #608 

If an admin role already exists, it will fail silently with the current methodology in the readme.  The new method will work for the assignment of the admin role whether it exists or not.  The seeds file is also updated, to change how admin roles are set.  It also eliminates a situation where the seed rake task errors out when the users already exist.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,47 +5,78 @@ class AddSeedObjects < ActiveRecord::Migration[5.1]
   password = Devise.friendly_token.first(8)
   puts "Password for all accounts: #{password}\n\n"
 
-  User.find_by_email('manydeposits@example.com').try(:destroy)
-  many_deposits = User.create(
-    email: 'manydeposits@example.com',
-    first_name: 'Many',
-    last_name: 'Deposits',
-    password: password,
-    password_confirmation: password,
-    ucdepartment: 'CCM Music')
-  puts "Account created: #{many_deposits.email}"
 
-  User.find_by_email('nodeposits@example.com').try(:destroy)
-  no_deposits = User.create(
-    email: 'nodeposits@example.com',
-    first_name: 'No',
-    last_name: 'Deposits',
-    password: password,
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -5,47 +5,81 @@ class AddSeedObjects < ActiveRecord::Migration[5.1]
...
-  User.find_by_email('manydeposits@example.com').try(:destroy)
-  many_deposits = User.create(
-    email: 'manydeposits@example.com',
-    first_name: 'Many',
-    last_name: 'Deposits',
-    password: password,
-    password_confirmation: password,
-    ucdepartment: 'CCM Music')
-  puts "Account created: #{many_deposits.email}"
+  if User.find_by_email('manydeposits@example.com')
```
hortongn: @chuckgreenman You should be able to refactor lines like

```
if User.find_by_email('manydeposits@example.com')
    many_deposits = User.find_by_email('manydeposits@example.com')
```
to
```
if many_deposits = User.find_by_email('manydeposits@example.com')
```

chuckgreenman: I made the change every where the pattern shows up in the seeds file.

PR #652, opened 2019-02-13
#630 Don't show metadata labels for empty fields.

### LINKED ISSUE #630: Some works show labels for metadata fields when those fields are empty
### Descriptive summary

If a work's metadata field is blank, that field should not be displayed on the work's show page.  Following the Scholar 4.0 deploy, however, there are some works that display field labels followed by nothing.

We need to make whatever changes are necessary to prevent labels for blank fields from being displayed.

NOTE: This **might** be related to a field having a `[]` vs. `""` vs. `nil` value.  It's possible that a field that has a blank string `""` as a value might be seen as a non-blank field.  Check the solr document (work.to_solr) to see if Solr thinks that some blank fields have content.

A Solr reindex on scholar.uc.edu **might** be needed after this is fixed.

If you cannot replicate this bug locally in your environment, we can do some testing on scholar.uc.edu to narrow down the cause.

Examples:

Blank Note field:
https://scholar.uc.edu/concern/images/kw52j9012

Blank Note and Genre field:
https://scholar.uc.edu/concern/images/73666561p?locale=en

Blank Required Software field:
https://scholar.uc.edu/concern/images/8g84mn28b

chuckgreenman:
Fixes #630  ;

Traditionally, blank fields are represented with `nil` in the work object.  Some blank fields were set to an array with an empty string as it's only member.  In ruby, `[""].blank?` => `false` so our traditional method of excluding them from the metadata presenter wasn't catching it.

To prevent showing empty metadata fields, I've added a function that test whether the values item it receives is just filled with empty strings.

A line has been added to the .rubocop.yml file instructing it to ignore the usage of html_safe, we're generating the HTML.  It's very possible that I'm missing something, but it doesn't look like we're doing any filtering when pulling user input from the db, should we rewrite that portion?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -217,6 +217,10 @@ RSpec/NamedSubject:
     - 'spec/services/hyrax/custom_stat_importer_spec.rb'
     - 'spec/models/collection_spec.rb'
 
+Style/FrozenStringLiteralComment:
+  Exclude:
+    - 'app/renderers/hyrax/renderers/attribute_renderer.rb'
+
 Style/ClassAndModuleChildren:
   Enabled: false
 
@@ -277,3 +281,4 @@ Lint/RescueException:
 Rails/OutputSafety:
   Exclude:
     - 'app/helpers/hyrax_helper.rb'
+    - 'app/renderers/hyrax/renderers/attribute_renderer.rb'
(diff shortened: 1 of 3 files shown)


chuckgreenman:
Ended up adding a line to the rubocop file about the frozen string literal comment.  Attribute renderer is basically a one giant unfrozen string manipulation.

PR #664, opened 2019-02-20
Make corrections to the collection export enhancements.

### LINKED ISSUE #662: More enhancements to collection exports
### Descriptive summary

1. There is one more place we need to re-enable the collection export link.  See where we commented it out here: https://github.com/uclibs/ucrate/commit/3c156ad25a472f7993c136cd176cdefbd59fac2e#diff-efd97eb4d0dc5d570e1d9e45e5115f61

1. On the Manage Collection Exports page, let's change the "Delete" column header to "Actions".  We might add additional actions to that column at some point.

1. On my machine, the timestamps for collection exports show a time that is 5 hours in the future.  So the time is in UTC instead of Eastern.  It's possible it won't show that way on a Mac host, but we need to convert it because scholar.uc.edu runs on Linux where the default time is UTC.

1. Add a column to the spreadsheet called "type" and for every row list the class of the object (e.g. GenericWork, Document, Collection, etc.).  Might be best to make this the very first column in the spreadsheet.

1. If the work has a DOI minted, we should be putting that into a "doi" column in the spreadsheet.  Whatever is in `work.doi` should go in the spreadsheet

1. Since Manage Collection Exports is now under the /dashboard/ route, we should display the dashboard sidebar when users are on that page (unless it is too much of a pain to do this)

1. ETDs save their publisher content to a `etd_publisher` field.  If possible, whatever is in `etd_publisher` should go in the spreadsheet under "publisher".  That is, both the `etd_publisher` and `publisher` fields should show up in the spreadsheet under a single "publisher" column.

chuckgreenman:
Fixes #662 ;

1. Re-enabled collection export link.
1. Changed delete column header on exports page to actions.
1. UCrate will now detect the timezone set on a users computer, set a cookie and then show the export in their timezone.
1. Added column to export spreadsheet for work type.
1. Added column to export spreadsheet for doi.
1. Collection exports page now has dashboard styling.
1. ETD publishers show up as publishers in the export spreadsheet.


config/metadata_export_map.yml
```diff
@@ -88,6 +92,7 @@ FileSet:
   - date_created
   - subject
   - language
+  - doi
```
hortongn: Remove the `doi` line since FileSets don't have that attribute

spec/fixtures/export.csv
```diff
@@ -1,12 +1,12 @@
...
+Medium	8	Medium	Putnam, Robert	test_email@example.com	Bowling Alone	All Rights Reserved	1	Arts and Letters	Social Science	Macmillan Reference	2000-01-01	Bowling Alone: The decline of American Social Capital			United States Politics	United States	Early Aughts	English	PDF	Medium Note	www.example.com/medium	00007				
+StudentWork	9	Student Work	Putnam, Robert	test_email@example.com	Bowling Alone	All Rights Reserved	1	Arts and Letters	Social Science	Macmillan Reference	2000-01-01	Bowling Alone: The decline of American Social Capital			United States Politics	United States	Early Aughts	English	PDF	Student Work Note	www.example.com/student_work	00008	Non Fiction	Kroll, Nick		Political Science
+Collection	10	Nested Collection title	Nested Collection creator	test_email@example.com	nested collection description		1																			
+Document	11	Nested Document	Putnam, Robert	test_email@example.com	Bowling Alone	All Rights Reserved	10	Arts and Letters	Social Science	Macmillan Reference	2000-01-01	Bowling Alone: The decline of American Social Capital			United States Politics	United States	Early Aughts	English	PDF	Document Note	www.example.com/document	00009	Non Fiction
```
hortongn: Let's add one more line to your fixture for a `FileSet` so that we test that file metadata is properly exported.

Then in your `spec/services/collection_metadata_csv_factory_spec.rb` I think you can just create a file set with a factory and then attach that file set to one of your existing work factories.  You can see an example of attaching a file to a work using factories here: spec/services/collection_metadata_csv_factory_spec.rb

spec/fixtures/export.csv
```diff
@@ -1,12 +1,15 @@
...
-11	Nested Document	Putnam, Robert	test_email@example.com	Bowling Alone	All Rights Reserved	10	Arts and Letters	Social Science	Macmillan Reference	2000-01-01	Bowling Alone: The decline of American Social Capital			United States Politics	United States	Early Aughts	English	PDF	Document Note	www.example.com/document	Non Fiction				
+type	id	title	creator	depositor	description	license	parent_id	college	department	publisher	date_created	alternate_title	journal_title	issn	subject	geo_subject	time_period	language	required_software	note	related_url	doi	genre	advisor	committee_member	degree
+Collection	1	Parent Collection		test_email@example.com	collection description																					
+Article	2	Article	Putnam, Robert	test_email@example.com	Bowling Alone Article	All Rights Reserved	1	Arts and Letters Article	Social Science	Macmillan Reference	2000-01-01	Bowling Alone: The decline of American Social Capital	American Governance	978-0-02-866249-7	United States Politics	United States	Early Aughts	English	PDF	Article Note	www.example.com/article	00001				
+FileSet	%{id_0}	FileSet Title	Greenman, Chuck	%{email_0}	FileSet description	All Rights Reserved	2
```
chuckgreenman: @hortongn I've addressed the change requests for this PR and modified the test file to use some string interpolations, let me know if you'd rather have it done some other way.

hortongn:
@chuckgreenman Both of the travis runs have spec/services/collection_metadata_csv_factory_spec.rb:289 failing

PR #667, opened 2019-02-22
#659 Generate all work types with content in all fields for testing

### LINKED ISSUE #659: Generate all work types with content in all fields for testing
### Descriptive summary

For testing and QA purposes, it would be nice if we could easily generate a work of every work type with content in all fields.

Update the `db:seed` task to also generate 8 works (one of each type)
* All fields should be completed.  The content in each field should make it obvious what that field is (e.g. for Time Period enter something like "time period")
* Multi-valued fields should have more than one value entered
* Have the manydeposits user own these works

The task should also create a collection with all metadata fields completed and the 8 works above should be added to that collection.

chuckgreenman:
Fixes #659  ;

This PR adds eight works with complete metadata added.  These works are assigned to the user manydeposits@example.com.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -104,4 +104,237 @@ class AddSeedObjects < ActiveRecord::Migration[5.1]
     puts "#{works_per_user} Generic Works created for #{user.email}"
   end
 
+  # Generate complete works
+
+  complete_collection = Collection.create(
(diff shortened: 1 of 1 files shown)


scherztc:
Merging with new issue for updates to Collection generated.  
https://github.com/uclibs/ucrate/issues/680

PR #697, opened 2019-03-15
Users shouldn't be able to self assign proxy rights

### LINKED ISSUE #675: Users should not be able to assign themselves as their own proxy
Moved from https://github.com/uclibs/scholar_uc_legacy/issues/1027

This may be a Hyrax bug (either documented or undocumented).

### Descriptive summary

I can assign myself as my own proxy.

### Expected behavior
I should not be able to find myself when searching for users to assign as a proxy, or be able to assign myself as my own proxy.  

### Actual behavior
I can assign myself as my own proxy.  When submitting a new work, I am then given a choice to submit on behalf of myself, or on behalf of my email address.

chuckgreenman:
Fixes #675  ;

We don't want users to be able to assign themselves proxy rights.  We now prevent them from doing so.  If a user attempts to add themselves as a proxy they will see an error message explaining that they can't add themselves as a proxy.

### DIFF
--- a/app/controllers/hyrax/depositors_controller.rb
+++ b/app/controllers/hyrax/depositors_controller.rb
@@ -5,12 +5,15 @@ module Hyrax
     def create
       grantor = authorize_and_return_grantor
       grantee = ::User.from_url_component(params[:grantee_id])
+
       if grantor.can_receive_deposits_from.include?(grantee)
         head :ok
-      else
+      elsif grantee.email != grantor.email
         grantor.can_receive_deposits_from << grantee
         send_proxy_depositor_added_messages(grantor, grantee)
         render json: { name: grantee.name, delete_path: sanitize_route_string(hyrax.user_depositor_path(grantor.user_key, grantee.user_key)) }
+      else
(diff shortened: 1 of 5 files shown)


app/controllers/hyrax/depositors_controller.rb
```diff
@@ -5,12 +5,15 @@ class DepositorsController < ApplicationController
...
+
       if grantor.can_receive_deposits_from.include?(grantee)
         head :ok
-      else
+      elsif grantee.email != current_user.email
         grantor.can_receive_deposits_from << grantee
         send_proxy_depositor_added_messages(grantor, grantee)
         render json: { name: grantee.name, delete_path: sanitize_route_string(hyrax.user_depositor_path(grantor.user_key, grantee.user_key)) }
+      else
+        render json: { error: "You can't be a proxy for yourself." }
```
hortongn: According to Coveralls, this line isn't covered by specs.

Add a context to depositors_controller_spec that verifies behavior when the user == the grantee.

PR #715, opened 2019-03-28
update change_manager gem version

### LINKED ISSUE #450: Emails are not being sent on removal for proxy and editor
While I was implementing the Change Manager created by Kyle Lawhorn, I noticed that emails were not being sent when you remove someone as a Proxy, or when you remove them as an editor. After consulting @hortongn about the issue, we checked and the issue is affecting production Scholar as well. 

~~Check with @hortongn about whether we still want an email for both proxy removal and editor removal.~~  We do want this.  Now that UCrate is in production, we should test on scholar.uc.edu to see what emails are not happening and fix it.

crowesn:
Fixes #450 ; refs https://github.com/uclibs/change_manager/pull/1

Change notification emails were not being consistently sent for proxy/editor removal. The change manager gem had some logic issues that have been fixed here: https://github.com/uclibs/change_manager/pull/1

This PR updates the Gemfile to point to the updated change_manager ref

* Update ref in Gemfile


hortongn:
@crowesn Sorry, but this now has a merge conflict.  Resolve and I'll merge.  I'll open a separate issue for fixing emails when an editor is removed.

crowesn:
@hortongn this should be good to go.

PR #723, opened 2019-04-03
Fix file labels in Create Work

### LINKED ISSUE #698: Specifying a display label during file upload has no effect
### Descriptive summary

When users are on a work's file upload tab, each file that is added offers the ability to set a "Display label".  That label should then be used to set the `title` for the fileset.  However, the custom labels appear to never be used.

### Expected behavior

Putting custom text in the "Display label" field should cause the fileset's title field to be set to that text after the work/files are saved.

If that behavior cannot be achieved, we should remove the "Display label" field from the file upload screen.

bsp3ars:
Fixes #698 

After looking into why the file labels work for batch upload but not file upload, it started to look like it was due to the different ways that the app/views/hyrax/uploads/_js_templates.html.erb file was called. SO rather than messing with that, I decided that it would be simplier to check whether the page was batch_uploads. If it was then show the label, else don't.

### DIFF
--- a/app/views/hyrax/uploads/_js_templates.html.erb
+++ b/app/views/hyrax/uploads/_js_templates.html.erb
@@ -65,12 +65,14 @@
                     </button>
                 </div>
             </div>
+            {% if (document.location["href"].includes("<%= hyrax.new_batch_upload_path(locale: nil) %>"))  { %}
             <div class="row">
                 <label for="title_{%=file.id%}" class="col-sm-2 control-label">Display label</label>
                 <div class="col-sm-10 padding-bottom">
                     <input type="text" class="form-control" name="title[{%=file.id%}]" id="title_{%=file.id%}" value="{%=file.name%}">
                 </div>
             </div>
+            {% } %}
         </td>
     </tr>
 {% } %}


app/views/hyrax/uploads/_js_templates.html.erb
```diff
@@ -65,12 +65,14 @@
                     </button>
                 </div>
             </div>
+            <% if current_page?('/batch_uploads/new') %>
```
hortongn: Instead of `'/batch_uploads/new'` let's use the route helper: `new_batch_upload_path(locale: nil)` in case the path for new batches ever changes.

hortongn: @bsp3ars Sorry, this isn't working as expected.  The conditional wants to check if the user is on the new batch upload page, but those pages have a URL like `/batch_uploads/new?locale=en&payload_concern=Article`.  So the payload_concern parameter is throwing it off.

Maybe it's easier to use `unless` instead of `if`?  Check that the user is not on the regular new work page?

bsp3ars:
@hortongn this is ready for a review

PR #733, opened 2019-04-19
Enable notifications for editor removal

### LINKED ISSUE #724: Emails are not being sent when editors are removed from works
### Descriptive summary

After #715 is merged, we will still have a bug where emails are not being sent when an editor is removed from a work.

The fix will likely need to take place in https://github.com/uclibs/change_manager

REF: https://github.com/uclibs/ucrate/pull/715

crowesn:
Fixes #724 

Enable email notifications for editor removal. Unfinished work was commented out--fixed the params and logic so that it works, removed skip on the specs.

* Add name param for permissions removal
* Uncomment notification
* Add spec


crowesn:
@hortongn I made changes as per our conversation - it should be good to go.

PR #735, opened 2019-04-23
Limit description to 5 lines on catalog index page.

### LINKED ISSUE #728: Only show a small amount of description field on /catalog page
### Descriptive summary

The catalog index page (/catalog) displays the entire description/abstract field for collections and works even if the field has tons of content.

catalog

1. We probably only want to show about five lines of the description, but the exact amount is up for discussion.  We should probably test with varying lengths and see what appeals to us the most.  If part of the description is hidden, we should indicate that with "..." at the end of the text.  It's possible this can be achieved using mostly/all CSS tweaks with something like `text-overflow: ellipsis;`.  If we can do it effectively using only CSS, that would be the best route.

2. If the column with of the description can be widened so it isn't so narrow, we should do that as well.

chuckgreenman:
Fixes #728 ; 

Descriptions on the catalog controller are now limited to five lines.  My approach was to use the Clamp.js library since there is no universally supported line clamping css approach.  Ended up adding class names to the facets on the catalog page, writing a function that selects all the elements with the description class and calls clamp.  If we decide that we want to change the number of lines we just change the number we pass clamp.

### DIFF
--- a/app/assets/javascripts/clamp.min.js
+++ b/app/assets/javascripts/clamp.min.js
@@ -1,0 +1,28 @@
+/*!
+* Clamp.js 0.5.1
+*
+* Copyright 2011-2013, Joseph Schmitt http://joe.sh
+* Released under the WTFPL license
+* http://sam.zoy.org/wtfpl/
+*/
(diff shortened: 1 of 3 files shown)


hortongn:
@chuckgreenman I tried this and it kinda works.  If I go strait to `http://localhost:3000/catalog` in my browser or reload the page, everything works well.  **But if I click the red [Browse] button the full descriptions are displayed**.  Once I refresh the page, they shorten.

(Full descriptions are also shown if I click the big [Browse] button on the home page)

Flaky issues with Javascript like that are typically caused by turbolinks.  It might be that a JS solution won't work for this unless we were to disable turbolinks in the right places.

(this was using the Chrome browser)

hortongn:
This may need to be rebased on top of the chromedriver fix.

hortongn:
@chuckgreenman In Chrome a slight bit of the 6th line of the description is shown.  See screenshot below.  Is there a way to compensate for this?  Firefox and Safari look fine.

Index Catalog    Scholar UC

chuckgreenman:
@hortongn I was able to reproduce this by changing the zoom around a bit, changed the padding css to margin, and it looks like it's working better now.

PR #757, opened 2019-06-17
Add link to parent work when viewing child works.

### LINKED ISSUE #710: Child works should link to their parent work
### Descriptive summary

A user viewing the public show page for a child work has no way of knowing there is a parent work or how to get to it.

For child works, we should include a link to the parent under the Relationships section (the same section that shows what collections a work belongs to).

The parent work can be found using `work.member_of`

Example:

### Relationships
**In Collection:**
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Global Marine Biodiversity Archive
**Attached to Parent Work:**
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Link to Parent Work

parent

scherztc:
Fixes #710 

* Add link to parent when viewing child work
* Add link to parent record
* Add return to parent record button on child work
* Create parent query service and test for service.

### DIFF
--- a/app/views/hyrax/base/_parent_relationship_table.html.erb
+++ b/app/views/hyrax/base/_parent_relationship_table.html.erb
@@ -1,0 +1,13 @@
+<% parent_solr_docs = ParentQueryService.query_parents_for_id(child_id) %>
+
+<dt> <%= t('hyrax.relationships_parent_row.label') unless parent_solr_docs.blank? %> </dt>
+  <dd>
+    <ul class="tabular">
+      <% parent_solr_docs.each do |item| unless parent_solr_docs.blank? %>
+        <li class='attribute attribute-title'>
+          <%= link_to item["title_tesim"].first, url_for_document(SolrDocument.find(item["id"])) %>
+        </li>
+      <% end %>
+     <% end %> 
+    </ul>
+  </dd>
--- a/app/views/hyrax/base/_relationships_parent_rows.html.erb
+++ b/app/views/hyrax/base/_relationships_parent_rows.html.erb
@@ -3,10 +3,13 @@
   <%= render 'relationships_parent_row', type: model_name, items: items, presenter: presenter %>
 <% end %>
 
+<%# [scholar-override] Show parent relationships for child %>
+<%= render 'parent_relationship_table', child_id: presenter.solr_document.id %>
+
+
 <%# Render grouped presenters. Show rows if there are any items of that type %>
 <% presenter.presenter_types.each do |type| %>
   <% presenter.grouped_presenters(filtered_by: type).each_pair do |_, items| %>
     <%= render 'relationships_parent_row', type: type, items: items, presenter: presenter %>
   <% end %>
 <% end %>
-
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -74,7 +74,11 @@ en:
         license: Rights
   hyrax:
     active_consent_to_agreement: I have read and agree to the
+    relationships_parent_row:
+      label: "In Parent Work:"
     base:
+      relationships_parent_row:
+        label: "In Collection:"
       form_progress:
         required_agreement: Check distribution license
       items:
--- a/app/services/parent_query_service.rb
+++ b/app/services/parent_query_service.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+module ParentQueryService
+  def self.query_parents_for_id(child_id)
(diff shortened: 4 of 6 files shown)


app/views/hyrax/base/_parent_relationship_table.html.erb
```diff
@@ -0,0 +1,16 @@
+<% parent_solr_docs = ParentQueryService.query_parents_for_id(child_id) %>
+
+<dt> <%= t('hyrax.relationships_parent_rows.label') %> </dt>
+<% if parent_solr_docs.blank? %>
+  <p><%= t('hyrax.relationships_parent_rows.empty_text') %></p>
```
hortongn: Instead of showing ""This work has no parents." just don't show anything when there is no parent (`unless parent_solr_docs.blank?`)

app/views/hyrax/base/_relationships_parent_rows.html.erb
```diff
@@ -3,10 +3,19 @@
...
+
 <%# Render grouped presenters. Show rows if there are any items of that type %>
 <% presenter.presenter_types.each do |type| %>
   <% presenter.grouped_presenters(filtered_by: type).each_pair do |_, items| %>
     <%= render 'relationships_parent_row', type: type, items: items, presenter: presenter %>
   <% end %>
 <% end %>
 
+<%# checks if user present and if they're an admin %>
+<% if current_user&.admin? %>
```
hortongn: Leave this out so we don't have "In admin set..." displayed

config/locales/hyrax.en.yml
```diff
@@ -73,6 +73,10 @@ en:
         description: Description
         license: Rights
   hyrax:
+
+    relationships_parent_rows:
+      empty_text: "This work has no parents."
+      label: "Parents:"
```
hortongn: Instead of "Parents:" say "In Parent Work:"

app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -0,0 +1,53 @@
+<%# [scholar-override] Show link to parent record %>
+<% parent_solr_docs = ParentQueryService.query_parents_for_id(presenter.solr_document.id) %>
+
+<div class="show-actions">
+
+  <% unless parent_solr_docs.blank? %>
+    <% parent_solr_docs.first(1).each do |item| %>
+      <%= link_to 'Back to Parent Record', url_for_document(SolrDocument.find(item["id"])), class: 'btn btn-success', data: { turbolinks: false } %>
```
hortongn: Get rid of the "Back to Parent Record" button.  Maybe that means you can get rid of this entire partial?

hortongn:
@scherztc Test notification

scherztc:
@hortongn :  This view test is getting extra complicated and is beginning to take up too much of my time. Neither Oregon State (https://github.com/osulp/Scholars-Archive/pull/1367/files) nor North Carolina (https://github.com/UNC-Libraries/hy-c/pull/108) worried about a view test.  I would like to roll back all the rails helper, controller helper, and the view test.  What do you think?

hortongn:
@scherztc Don't you already have the view spec working?

scherztc:
No.  Not while keeping the correct labels for both Parents and Collections.  Hyrax had these two things grouped together initially for these views.  It is an inherited pain point and probably part of why they lost this functionality in the beginning.

hortongn:
Ah, gotcha.  I think the fact that "In Collections" silently broke in your last commit is exactly why we do need some kind of testing in place.  You accidentally broke the collections label while working on the parent work label.  Ideally some test should have failed and alerted us.

How about instead of a spec we verify it works in the manual test script?  So remove all of the view spec stuff and leave just the service spec.  Then also as part of your PR add a piece to the test_script.md under Section 2 that 

1. creates a child work (you can use the work already in the script as the parent)
1. adds the child work to a collection
1. asks the tester to visit the child work's show page to verify that In Collections and In Parent Work are displaying properly.

scherztc:
Sounds good.  I am also completely perplexed on how this test is passing in Hyrax.

https://github.com/samvera/hyrax/blob/master/spec/views/hyrax/base/_relationships.html.erb_spec.rb

hortongn:
My guess is the view itself in Hyrax works, but this feature is maybe broken elsewhere in Hyrax.  So the view spec passes by stubbing what it needs, but there may be a problem in a Hyrax controller, presenter, etc. that breaks this feature in real life.

That's one of the reasons why feature specs can be helpful in an app.  It's possible that a view, a controller, and a model all have their own specs and individually appear to work, but once they start trying to talk to each other there can be bugs that only a feature spec would catch.

PR #766, opened 2019-07-23
Fix auto-fill for the publisher field

bsp3ars:
After troubleshooting with @chuckgreenman, the issue seemed to come from the fact that the setPublisher file wasn't being loaded when the user tried to edit the file. This was due to the way that it used to load via turbolinks. This was changed to whenever a page loaded, and voila.

### DIFF
--- a/app/assets/javascripts/set_publisher.js
+++ b/app/assets/javascripts/set_publisher.js
@@ -27,5 +27,6 @@ var setPublisher = function()   {
         }
     }
 }
-
-$(document).on('turbolinks:load', setPublisher)
+document.addEventListener("turbolinks:load", function() {
+    setPublisher();
+})


scherztc:
With my first look at it this:
1.  When creating a work with no publisher, I get a blank flash warning and instructions above the blank publisher field.  I then entered University of Cincinnati as the Publisher and attempted to submit work and I get an Unknown agent type ""n error.

scherztc:
2.  Seeing same bug on EDIT.

chuckgreenman:
Here's a list of turbolinks events that may come in handy: https://github.com/turbolinks/turbolinks/blob/master/README.md#full-list-of-events

scherztc:
Here are some notes about working with Turbolinks and Rails 5: https://stackoverflow.com/questions/36110789/rails-5-how-to-use-document-ready-with-turbo-links

bsp3ars:
Ready for another review @scherztc

Issue #771, opened 2019-08-05
setCreator is throwing an error on every page that doesn't have a current_user.


scherztc:
### Descriptive summary

setCreator is getting called on every page due to the way we have it defined with Turbolinks.  Many of the pages don't have a current_user so it throughs an error in the java console.

<img width="1276" alt="Screen Shot 2019-08-05 at 3 57 18 PM" src="https://user-images.githubusercontent.com/5084562/62491567-4d39f000-b79a-11e9-80b6-44aa49ac701f.png">

### Expected behavior

That setCreator only load on the proper page or that setCreator has a conditional for current_user

### Steps to reproduce the behavior

1. Visit Home Page
1. Open Java Console in Developer Tools

### Related work

Link to related issues or prior related work here.

PR #784, opened 2019-08-21
Fix auto-populate creator

### LINKED ISSUE #767: New Collection does not auto-populate Creator field
### Descriptive summary

As a user when I create a new collection from the dashboad or navigation menu, the creator is blank.  

### Expected behavior

It is expected that this field be populated by last name, first name when the form is loaded.

1. Create New Collection.
1. Examine the Creator field in the form.

bsp3ars:
Fixes #767

Fixes our issue where the setCurrentUserAsCreator function was throwing an error on the collection page because of a missing div. It appears that initially we didn't want to have it auto-populate. I kept the etd_creator field as not auto-populating since that seems to be the current use case.


bsp3ars:
I was also able to fix #771 by changing setCurrentUserCollegeAsCollege.

crowesn:
@bsp3ars Maybe tack on a quick check for this functionality on the collection feature spec?

spec/features/hyrax/dashboard/collection_spec.rb
```diff
@@ -262,6 +262,7 @@
         title_element.set("Test Collection") # Add whitespace to test it getting removed
 
         creator_element = find_by_id("collection_creator")
+        expect(creator_element.value).to eq("User, Sample")
```
bsp3ars: @crowesn Added check for the users name before we change it

PR #823, opened 2019-10-31
Add research directory to user profile

### LINKED ISSUE #731: User profile pages should include a link to the user's Research Directory profile
(moved from https://github.com/uclibs/scholar_uc_legacy/issues/18)

### Descriptive summary

Investigate a reliable way to link to a user's Research Directory profile from their Scholar profile page.

Example:
On https://scholar.uc.edu/users/hortongn@ucmail-dot-uc-dot-edu?locale=en, include a link in the sidebar called "Research Directory profile" that links to https://researchdirectory.uc.edu/p/hortongn

The format seems to be `https://researchdirectory.uc.edu/p/<sixplus2>`, but we should not assume that a page in RD exists for everyone.  So we should have a method to verify the link is resolvable before we display it.

We can probably get the user's sixplus2 from their `user.uid` field.

bsp3ars:
Fixes #731

Adds a link to a persons research directory page to their user profile. Note: This includes a migration that adds a string to the User db for a link to their page


crowesn:
I also noticed that the research directory link has target=blank while the webpage link does not open a new window. We should shoot for parity?

PR #824, opened 2019-11-01
Add more fields to the works report.

### LINKED ISSUE #822: Add remaining attributes to manifest
REF: https://github.com/uclibs/ucrate/issues/673#issuecomment-545456654

The manifest currently doesn't include all attributes when it's generated.  We should add the fields that are missing.

scherztc:
Fixes #822

Adds depositor, on_behalf_of, and other fields to work manifest report.

### DIFF
--- a/app/services/works_report.rb
+++ b/app/services/works_report.rb
@@ -35,10 +35,12 @@ class WorksReport < Report
       alt_date_created
       alt_description
       alternate_title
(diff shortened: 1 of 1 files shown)


hortongn:
Before we merge this, can we manually add the fields to production and generate a report to see if it satisfies James's needs?

PR #847, opened 2020-01-14
Adds a one worker limit to the fixity_check sidekiq queue.

### LINKED ISSUE #835: Run fixity checks in a dedicated sidekiq queue
### Descriptive summary

Following the work in #727, we need to run fixity check jobs in a new "fixity" sidekiq queue.  The queue should be limited to 1 concurrent job since the fixity checks don't behave when multiple checks are running at the same time. 

If we cannot accomplish this with our current free version of sidekiq, we will need to discuss options.

Resources:
* https://github.com/mperham/sidekiq/wiki/Advanced-Options

As a final task for this issue, we need to set up a cron job to run the fixity checks regularly.  Need to discuss how often.

scherztc:
Fixes #835 

This PR uses a sidekiq_fetch_limit job to limit the fixity_check queue to 1 worker.
We then had to bring in app/jobs/ overwrite from Hyrax to assign it to the fixity_check queue.
We then included a mock script to be used with cron.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ end
 gem 'browse-everything', github: 'uclibs/browse-everything', branch: 'master'
 gem 'hydra-remote_identifier', github: 'uclibs/hydra-remote_identifier', branch: 'scholar-datacite'
 gem 'kaltura', '0.1.1'
+gem 'sidekiq-limit_fetch'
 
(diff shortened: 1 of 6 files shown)


scherztc:
I tested this with 50 files and then were added and characterized with the default queue and then the Fixity Check job uses the fixity_check queue.

PR #864, opened 2020-03-16
[WIP] Hide Relationship if empty

### LINKED ISSUE #805: Hide Relationships if empty
We currently do not hide the Relationships section of a work if it is empty, and this makes it somewhat confusing especially if there is a child work and it doesn't show up there. If we change the partial to hide if empty, I believe that it would be more intuitive and easier to understand for the end user.
Screenshot_2019-09-03 Test

Uzval-Mallepeddi:
Fixes #805 

Hide the relationships tab if there isn't any relationship with the work. Relationship includes parent work and collection.

* Hide the relationship tab

Blocker
- In the specs of _relationship partial, i am unable to use the method of an instance variable. "presenter" is the instance variable. "presenter_types" is the method of the instance variable. When i run the specs, it results in following error: 

Failure/Error: <% presenter.grouped_presenters(except: @presenter.presenter_types).each_pair do |model_name, items| %>
     
     ActionView::Template::Error:
       undefined method `presenter_types' for nil:NilClass


hortongn:
@Uzval-Mallepeddi The method is undefined because `presenter` is nil.  It's not defined.  You need a `let` statement that creates the presenter object.  In your case, a `WorkShowPresenter`.

Uzval-Mallepeddi:
@hortongn I've declared the "presenter" variable in `_relationship` partial and `show` spec with the `let` statement. I've used the same approach as specified in the hyrax link you shared.

`let(:presenter) do
    Hyrax::WorkShowPresenter.new(work_solr_document, ability, request)
  end`

PR #866, opened 2020-03-25
Move Collection Information

### LINKED ISSUE #807: Move Collection information
For both the dashboard and the public view of collections, the Parent Collections section seems to be in the wrong spot. This should be as simple as moving the partial to the correct spot. I included pictures to show where I believe they should go.
Screenshot_2019-09-03 Test 2 Scholar UC
Screenshot_2019-09-03 Test 2 Scholar UC(1)

Uzval-Mallepeddi:
Fixes #807 

- Re-organizes the "Parent Collection" information to be displayed after the search form. (If Parent Collection exists). 
- The re-organized structure is as follows:
            - Search Form
            - Parent Collection
            - Sub Collection
            - Works

### DIFF
--- a/app/views/hyrax/collections/show.html.erb
+++ b/app/views/hyrax/collections/show.html.erb
@@ -56,17 +56,6 @@
     <div class="col-md-8 hyc-description">
       <%= render 'collection_description', presenter: @presenter %>
 
-      <% if @presenter.collection_type_is_nestable? && @presenter.total_parent_collections > 0 %>
-          <div class="hyc-blacklight hyc-bl-title">
-            <h2>
-              <%= t('.parent_collection_header') %> (<%= @presenter.total_parent_collections %>)
-            </h2>
-          </div>
-          <div class="hyc-blacklight hyc-bl-results">
-            <%= render 'show_parent_collections', presenter: @presenter %>
-          </div>
-      <% end %>
-
     </div>
     <div class="col-md-4 hyc-metadata">
       <% unless has_collection_search_parameters? %>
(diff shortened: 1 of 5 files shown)


spec/features/hyrax/batch_edit_spec.rb
```diff
@@ -82,10 +82,10 @@
       click_on 'batch-edit'
       find('#edit_permissions_link').click
       expect(page).to have_content('Batch Edit Descriptions')
-
       # Set visibility to private
       within "#form_permissions_visibility" do
         batch_edit_expand('permissions_visibility')
+        find('#expand_link_permissions_visibility').click
```
hortongn: Is this needed?  Was the spec not passing?

spec/views/hyrax/dashboard/collections/show.html.erb_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
```
hortongn: The spec will run fine if you run the entire test suite, but if we run just this file it will error because it needs a `require rails_helper` at the top.  Any time you bring a spec over from Hyrax, you need to require rails_helper.

app/views/hyrax/collections/show.html.erb
```diff
@@ -94,6 +83,18 @@
     </div>
   </div>
 
+  <!-- Parent Collections -->
```
hortongn: If you are changing app/views/hyrax/collections/show.html.erb then you need to add a view test.  Bring over spec/views/hyrax/collections/show.html.erb_spec.rb from Hyrax and add the same 'when the rendered collection has parent collection' test from your other view spec.

spec/features/hyrax/batch_edit_spec.rb
Uzval-Mallepeddi: Yes. The test suite is failing by saying that Capybara cannot find the "#generic_work_visibility_authenticated" is not found. When i tested it via DOM, looks like the element is hidden. So, i made sure to click the parent of it to make the element visible.

Uzval-Mallepeddi:
> @Uzval-Mallepeddi Great job. The behavior is perfect. I've requested a couple of spec changes. So see the inline comments and let me know if you have questions.

> @hortongn Awesome! I'll let you know after making changes.

Uzval-Mallepeddi:
@hortongn I'll let you know after making changes.

Uzval-Mallepeddi:
@hortongn I think this should be good to go.

PR #868, opened 2020-03-28
Hide Relationships if empty

### LINKED ISSUE #805: Hide Relationships if empty

Uzval-Mallepeddi:
Fixes #805  

- Hides the "Relationships" header if the collection/work doesn't have any parent, sub-collection, child-work. Displays the "Relationships" header if there are any parent/child.

- Specs written only for `show.html.erb_spec.rb`. Since `show.html.erb` file calls `_relationships.html.erb` partial which conditionally calls both `_relationships_parent_rows.html.erb` and `_parent_relationship_table.html.erb`

### DIFF
--- a/app/views/hyrax/base/_parent_relationship_table.html.erb
+++ b/app/views/hyrax/base/_parent_relationship_table.html.erb
@@ -1,5 +1,5 @@
 <% parent_solr_docs = ParentQueryService.query_parents_for_id(child_id) %>
-
+<h2><%= t('hyrax.base.relationships.label') unless parent_solr_docs.blank? %></h2>
 <dt> <%= t('hyrax.relationships_parent_row.label') unless parent_solr_docs.blank? %> </dt>
   <dd>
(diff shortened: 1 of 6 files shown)


scherztc:
@Uzval-Mallepeddi When a work has both a parent work and is part of a collection we get the Relationship label twice.   I checked on production scholar and we should only be getting this label once.

Screen Shot 2020-04-06 at 5 04 16 PM

scherztc:
Screen Shot 2020-04-06 at 5 12 16 PM

Uzval-Mallepeddi:
@scherztc This should do it. Could you please review and let me know if that's good ?

PR #871, opened 2020-04-03
[WIP] Highlight Search terms

### LINKED ISSUE #705: Highlight search terms on catalog view page
(moved from https://github.com/uclibs/scholar_uc_legacy/issues/222)

### Descriptive summary

When the user's search terms appear in the the search results, highlight them so the user can more easily see the matches.

Example: if the user searches for `library degree`, highlight any words in the search results that caused those records to be included.  See example below.

highlight

Uzval-Mallepeddi:
Fixes #705 

[WIP] Just to make sure if this is exact functionality that this feature needs.

### DIFF
--- a/app/views/catalog/_index_header_list_collection.html.erb
+++ b/app/views/catalog/_index_header_list_collection.html.erb
@@ -1,0 +1,21 @@
+<div class="search-results-title-row">
+  <% vars = request.query_parameters["q"].to_s.downcase %>
+  <% values = vars.split(" ") %>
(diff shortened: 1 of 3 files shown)


hortongn:
@Uzval-Mallepeddi We only want to highlight the individual words from the search box.  Your solution is highlighting the entire title or entire description field.

If I search for `library degree` it should only highlight the words "library" and "degree" in the titles and descriptions.

See the example image listed in the issue:

highlight

PR #873, opened 2020-04-16
[WIP] Highlight search terms

Uzval-Mallepeddi:
WIP

- When a run a spec on 'spec/views/catalog/_index_header_list_default.html.erb_spec.rb', it gives me an error `ActionView::Template::Error:
       undefined method 'constantize' for nil:NilClass`. I tried to debug and look at the instance variables set. Looks like everything is set to their respective values. Not sure of what the problem was.

### DIFF
--- a/app/helpers/mark_helper.rb
+++ b/app/helpers/mark_helper.rb
@@ -4,13 +4,13 @@ module MarkHelper
   def catalog(input_text)
     value = input_text.to_s
     filter_chars = params[:q].to_s.split(' ')
-    if !filter_chars.empty?
+    if filter_chars.empty?
+      value = value.to_s
+    else
       filter_chars.each do |char|
-        value = value.gsub!(/#{char}/i, '<mark>' + char.capitalize + '</mark>').to_s
+        value = value.gsub!(/#{char}/i, '<mark>' + char + '</mark>').to_s
         value = input_text.to_s if value.empty?
       end
-    else
-      value = value.to_s
     end
 
     value.html_safe
--- a/app/helpers/mark_helper.rb
+++ b/app/helpers/mark_helper.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module MarkHelper
+  def catalog(input_text)
+    value = input_text.to_s
+    filter_chars = params[:q].to_s.split(' ')
+    if !filter_chars.empty?
+      filter_chars.each do |char|
+        value = value.gsub!(/#{char}/i, '<mark>' + char.capitalize + '</mark>').to_s
+        value = input_text.to_s if value.empty?
+      end
+    else
+      value = value.to_s
+    end
+
+    value.html_safe
+  end
+end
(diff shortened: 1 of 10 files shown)


spec/views/catalog/_index_header_list_default.html.erb_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'spec_helper'
+
+RSpec.describe 'catalog/_index_list_default', type: :view do
+  let(:attributes) do
+    { 'id' => '123',
+      'title_tesim' => 'Test Work',
```
hortongn, line 9: @Uzval-Mallepeddi So when I run the spec I get the error you mentioned:

```
ActionView::Template::Error:
       undefined method `constantize' for nil:NilClass
     # /home/vagrant/.rvm/gems/ruby-2.5.5@ucrate/gems/hyrax-2.6.0/app/models/concerns/hyrax/solr_document_behavior.rb:74:in `hydra_model'
     # /home/vagrant/.rvm/gems/ruby-2.5.5@ucrate/gems/hyrax-2.6.0/app/models/concerns/hyrax/solr_document_behavior.rb:61:in `to_model'
     # ./app/views/catalog/_index_header_list_default.html.erb:2:in `_app_views_catalog__index_header_list_default_html_erb___3213301723848490191_88284520'
```
The important thing to look at is the first line of the trace under the error:
`app/models/concerns/hyrax/solr_document_behavior.rb:74:in 'hydra_model'`
This tells you the error is being thrown by line 74 in solr_document_behavior.rb.  This is where you can put your byebug if you wanted to examine it.

Line 74 is this: `first(Solrizer.solr_name('has_model', :symbol)).constantize`

So "undefined method `constantize' for nil:NilClass"` means that `first(Solrizer.solr_name('has_model', :symbol))` is returning nil instead of the name of a class (in this case a work type model).  Solrizer is trying to look up the value of `has_model` in your solr document, but your spec doesn't include `has_model` in your `let(:attributes)` statement.

I think if you add something like `"has_model_ssim" => ["GenericWork"],` to your `let(:attributes)` statement, it will either fix the spec or at least get you to the next error.

app/helpers/mark_helper.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module MarkHelper
+  def catalog(input_text)
+    value = input_text.to_s
+    filter_chars = params[:q].to_s.split(' ')
+    if !filter_chars.empty?
+      filter_chars.each do |char|
+        value = value.gsub!(/#{char}/i, '<mark>' + char.capitalize + '</mark>').to_s
```
hortongn, line 9: I don't think we want `char.capitalize`.  We want to highlight the word, but not capitalize it.

Also, the default highlight color is a little hard to see.  Can you use CSS to change the <mark> tag to a brighter yellow?  Maybe #ffff00

app/helpers/mark_helper.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module MarkHelper
+  def catalog(input_text)
+    value = input_text.to_s
+    filter_chars = params[:q].to_s.split(' ')
+    if !filter_chars.empty?
```
hortongn, line 7: In ruby instead of using `if !` it's nicer to use `unless filter_chars.empty?`

PR #878, opened 2020-06-02
WIP : Configures CircleCI

scherztc:
Fixes #issuenumber ; refs #issuenumber

More detailed description, if necessary. Try to be as descriptive as you can: even if you think that the PR content is obvious, it may not be obvious to others. Include tracebacks if helpful, and be sure to call out any bits of the PR that may be work-in-progress.

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

* 
* 
*


.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+        default: '5.1.6'
+      solr_config_path:
+        type: string
+        default: 'solr/config'
+      solr_core:
+        type: string
+        default: 'hydra-test'
+      bundler_version:
+        type: string
+        default: '2.0.1'
```
hortongn, line 43: @scherztc Try changing the bundler_version to `1.17.3`.  We've had problems in the past with using bundler 2.x.  That may relate to your db:migrate problem.

Issue #739, opened 2019-05-07
Switch from Travis to CircleCI


hortongn:
### Descriptive summary

The Samvera community is moving their repos from Travis to CircleCI.  We should research service and switch Scholar to it.

We only want to switch right now if we can use CircleCI with Scholar for free.

There are many examples of other repos switching to CircleCI:
* https://github.com/samvera/solrizer/commit/fad269d496605082c6e4bbc0b1080ad25facad04
* https://github.com/samvera/ldp/commit/14630f508294e4c9522bb7170a299f88d5713d07#diff-1d37e48f9ceff6d8030570cd36286a61
* https://github.com/samvera/hydra-pcdm/commit/fff8a8d594e121aa068da0455a981b1d2a89d5c8
* https://github.com/samvera/hyrax/commit/408f95ea3baeca39f5178664185caf4ba12de525#diff-1d37e48f9ceff6d8030570cd36286a61

Look into Samvera's CircleCI Orb (See some of the commits above)

Recording of Travis to CircleCI discussion (for Hyku):
https://iu.mediaspace.kaltura.com/media/CircleCI+knowledge+share+and+Hyku+swarm+-+May+2nd%2C+2019/0_bzw7tbz2

scherztc:
Samveras CirclCi Orb : https://circleci.com/orbs/registry/orb/samvera/circleci-orb

scherztc:
https://github.com/samvera-labs/samvera-circleci-orb

scherztc:
This job has been blocked because resource-class:large is not available on your plan. Please upgrade to continue building.

Issue #494, opened 2018-11-05
Resolve Travis / Rspec Issues


bsp3ars:
This issue will contain a listing of all of the known issues that are consistently appearing in Travis / Rspec. 

- [ ] `Create a work a logged in user allows fields to end with a double quote`
     `Failure/Error: select 'In Copyright', from: "image_rights_statement"`
     
     `Capybara::ElementNotFound:`
       `Unable to find visible select box "image_rights_statement" that is not disabled`

- [ ] `Selenium::WebDriver::Error::NoSuchDriverError:`
              `invalid session id`

- [ ] It appears that when Travis runs, it does not always run every test due to the fact that we have had to do several cleanups for spec tests that should have failed but didn't

- [x] Something to consider is to update the coveralls gem to see if that fixes any miscellaneous bugs

PR #881, opened 2020-06-24
Configures CircleCI.

scherztc:
Fixes Issue #739 
Fixes issue #494

Uses Samvera ORB
Pins Puma gem
Pins Rake gem


scherztc:
We are green again and coverage is updated, BUT still low.  I wonder if this number represents only job 141.2

Screen Shot 2020-07-07 at 5 32 38 PM

scherztc:
After using explicit rspec commands instead of rspec_parallel from Samvera

Screen Shot 2020-07-08 at 12 35 20 AM

.circleci/config.yml
```diff
@@ -0,0 +1,104 @@
...
+#                  bundle exec rspec \
+#                    --format progress \
+#                    --out /tmp/test-results/rspec.xml \
+#                    --format progress \
+#                  $TEST_FILES
+
+            - samvera/parallel_rspec
+
+            - store_artifacts:
+                path: tmp/screenshots
```
hortongn, line 94: @scherztc See the screenshot below.  I looked closer at the circleci interface and can see the total number fo tests and how to view each parallel build.

build

I added the coveralls token to the circleci account and it likes it.  But I think we have to change `path: tmp/screenshots` to `path: coverage` to get the coveralls integration working properly.  So try that change and clean up the rest of the file so we can merge.  If we get coveralls working here we can try the same on our other rails projects.

README.md
```diff
@@ -1,5 +1,7 @@
 # Scholar@UC
 ![Travis (.org) branch](https://img.shields.io/travis/uclibs/ucrate/develop.svg) ![Coveralls coverage](https://img.shields.io/coveralls/github/uclibs/ucrate/develop.svg)
+[![CircleCI](https://circleci.com/gh/uclibs/ucrate.svg?style=svg)](https://circleci.com/gh/uclibs/ucrate)
+[![Coverage Status](https://coveralls.io/repos/github/uclibs/ucrate/badge.svg?branch=master)](https://coveralls.io/github/uclibs/ucrate?branch=master)
```
hortongn, line 4: Also be sure to update the README docs to use 8080 port for fedora testing

PR #896, opened 2020-10-13
Fix/my sql

scherztc:
Fixes MySQL and Capistrano problem for migrations


hortongn (changes requested):
@scherztc Please squash the commits

hortongn (approved):

PR #898, opened 2020-11-09
Restricts access to BulkRAX to admins.

scherztc:
Fixes: Limiting BulkRAX to only admin users.

We needed to limit BulkRAX routes and dashboard options to only admin users.  We did this with a read access check for the admin dashboard.


scherztc:
When I ported the importer test over from BulkRAX, some of the test are failing.  This could either be related to the code changes we made since they were permission based or could have to do with these same permissions being set in a before hook during the test or the problems could still be around factories used in the test.  I have commented out all of the failing test.  Any feedback on them would be appreciated.

hortongn (commented):
@scherztc I tried reopening the ExportersController instead of dull overwriting it and it seems to work.  No need to override all of the methods.  Take a look at code below.  You shouldn't have to test the whole controller this way.

```
require_dependency "bulkrax/application_controller"
gem_dir = Gem::Specification.find_by_name("bulkrax").gem_dir
require "#{gem_dir}/app/controllers/bulkrax/exporters_controller.rb"

module Bulkrax
  class ExportersController < ApplicationController
    include Hyrax::ThemedLayoutController
    include Bulkrax::DownloadBehavior
    before_action :ensure_admin!
    before_action :authenticate_user!
    before_action :set_exporter, only: [:show, :edit, :update, :destroy]
    with_themed_layout 'dashboard'

  private

    def ensure_admin!
      authorize! :read, :admin_dashboard
    end
  end
end
```

hortongn (approved):
@scherztc 👍

PR #902, opened 2020-12-21
Adds FileSystem to BrowseEverything

scherztc:
Adds the mounted server path as an option for BrowseEverything.

This is useful for Content Services to upload large files into a shared Isilon mount and then the users will easily be able to attach the large files to their own works. 

No need for ownership changes.


hortongn (approved):
@scherztc This is a simple and elegant solution until we get OneDrive integrated.

PR #915, opened 2021-03-12
Uses environment variable for host name to allow for logs on multiple…

### LINKED ISSUE #870: Capistrano not linking logs correctly
### Descriptive summary

The capistrano start_qa.sh and start_prod.sh script are only setting up logging for one of the load balanced web servers.  

Both scripts only set up log links for libschqwl1 and libschpwl1.  The libschqwl2 and libschpwl2 servers are left out:
https://github.com/uclibs/ucrate/blob/develop/script/start_qa.sh#L5-L11

An easy fix might be to use the $HOSTNAME environment variable in the scripts instead of hard coding the server name.  That way the same script will use the proper server name when run on different servers.

scherztc:
… servers.

Fixes #870

We will need to update all of the .env files on prod and qa (4 of them).

This should be tested on DEV.

### DIFF
--- a/app/models/hyrax/virus_scanner.rb
+++ b/app/models/hyrax/virus_scanner.rb
@@ -67,8 +67,8 @@ module Hyrax
 
     private
 
-      def warning(msg)
-        ActiveFedora::Base.logger&.warn(msg)
-      end
+    def warning(msg)
+      ActiveFedora::Base.logger&.warn(msg)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


hortongn (changes requested):
@scherztc I don't think the variable approach will work.  However, we can use the `hostname` command on each server to get the current hostname.  So something like this should work:

    touch /mnt/common/scholar-logs/`hostname`_production.log

That should produce **touch /mnt/common/scholar-logs/libschqwl1_production.log** and **touch /mnt/common/scholar-logs/libschqwl2_production.log** on each respective server.

Also, remove the change made to SCHOLAR_DOI_SHOULDER in .env.development

hortongn (changes requested):
@scherztc Changes look good, but it has a conflict because of the robots.txt change.  Please rebase.

hortongn (approved):

PR #1095, opened 2023-11-01
1082 - Remove stub chain

### LINKED ISSUE #1082: Using 'stub_chain' and should
### Descriptive summary

Using `stub_chain` from rspec-mocks' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` instead. Called from /Users/scherztc/Workspaces/ucrate/spec/views/catalog/_index_list_default.html.erb_spec.rb:47:in `block (3 levels) in <top (required)>'.

1. Do this
1. Then do this...

### Related work

Janell-Huyck:
Fixes #1082

**_Removes old stub_chain methods_**

### Summary
This PR refactors the test suites for **_user_util_links.html.erb_spec.rb** and **_index_list_default.html.erb_spec.rb** to remove the deprecated stub_chain methods. The code now relies on cleaner, up-to-date RSpec syntax, and is more maintainable as a result.

### Details

- Deprecated stub_chain methods were removed from both files, replacing them with cleaner helper methods to improve readability and maintainability.
- The test in _index_list_default.html.erb_spec.rb was making a direct Solr call, which has now been mocked to isolate the test from external dependencies.  This also allows the test to be run outside of the Solr environment.  (If the stub is removed and the Solr environment is active, the test still passes.)

##### Helper method for setting up a mock for current_ability
```
def stub_current_ability(return_value)
  current_ability = double('CurrentAbility')
  allow(view).to receive(:current_ability).and_return(current_ability)
  allow(current_ability).to receive(:can_create_any_work?).and_return(return_value)
end
```

##### Usage inside the before block
```
stub_current_ability(true)
```
### Why "allow" Over "expect"
We opted to use allow instead of expect because these stubs were set up in before-do blocks and are not the primary behavior under test. By using allow, we can establish the necessary background for the actual tests without enforcing strict expectations on the messages received by our doubles, making the tests less brittle and more focused on what we are actually interested in testing.

### Additional changes
- added two tests for inheritance to offset supposed decrease in test coverage
- added option to pull environment variable for FITS_FILENAME so that developers using "fits" instead of "fits.sh" will be able to add that ENV variable to their local config and override the default "fits.sh".  If they use a local environment variable, it will eliminate accidental overwrites of the `config/initializers/hyrax.rb` file with local changes.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -28,7 +28,7 @@ ActiveRecord::Schema.define(version: 2022_06_09_001128) do
     t.string "identifier"
     t.string "collection_ids"
     t.string "type"
-    t.integer "importerexporter_id"
+    t.integer "importerexporter_id", null: false
     t.text "raw_metadata", limit: 16777215
     t.text "parsed_metadata", limit: 16777215
     t.datetime "created_at", null: false
@@ -37,6 +37,7 @@ ActiveRecord::Schema.define(version: 2022_06_09_001128) do
     t.datetime "last_succeeded_at"
     t.string "importerexporter_type", default: "Bulkrax::Importer"
     t.integer "import_attempts", default: 0
+    t.index ["importerexporter_id"], name: "index_bulkrax_entries_on_importerexporter_id"
   end
 
   create_table "bulkrax_exporter_runs", force: :cascade do |t|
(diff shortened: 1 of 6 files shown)


scherztc (changes requested):
Remove this. It doesn’t have anything to do with the PR and am not sure I want this uncommented by default

db/schema.rb
```diff
@@ -28,7 +28,7 @@
     t.string "identifier"
     t.string "collection_ids"
     t.string "type"
-    t.integer "importerexporter_id", null: false
+    t.integer "importerexporter_id"
```
scherztc: Remove these changes.  They aren’t relevant to this PR.  I don’t want to go making any changes to the db/schema by hand.

scherztc (approved):

PR #1106, opened 2023-11-06
Allow for env variable for fits in hyrax config

Janell-Huyck:
Adds an optional development environment variable "FITS_FILENAME" to hyrax.rb config file

Developers who have installed "fits" via homebrew have the fits file stored in their computer as "fits", while developers who have installed "fits" via Harvard have the fits file stored in their computer as "fits.sh".  The app defaults to "fits.sh", which causes it to crash when run locally from a "fits" machine.  The workaround to this has been to update the hyrax config file to indicate that the path should be "fits" instead of the default "fits.sh".

This PR adds an optional environment variable for "fits" developers, so that they can put `FITS_FILENAME` in their .env.development.local file to override the app's location of fits.  

Developers who had their app working correctly with the "fits.sh" default do not need to put in the new environment variable, and should notice no change in functionality.

Also adds note in README about the FITS_FILENAME variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@
 1. Run brakeman: ```bundle exec brakeman -q -w 2```
 1. Run bundler-audit: ```bundle-audit check --update```
 
+## Notes for local development
+1. If you installed fits on a mac via homebrew, you may be getting an error saying that there is no fits.sh file.  If this happens, add the following to your .env.development.local: `FITS_FILENAME=fits`
+
 ## Project Samvera
 This software has been developed by and is brought to you by the Samvera community. Learn more at the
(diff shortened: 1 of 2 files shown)


scherztc (commented):
Do not UNCOMMENT this.  FITS is working fine in all of our server environments.  I really don't want to make a change to a working environment for a local configuration.

We are also trending towards FITS as a servlet in our server environments.  Maybe we can wait to address this when we address that : https://github.com/harvard-lts/FITSservlet

Janell-Huyck:
Issue fixed locally by modifying .zshrc file, adding `alias fits.sh='fits'`

Issue #985, opened 2022-04-13
Deprecation Warning : sqlite3_adapter.rb


scherztc:
/Users/scherztc/.rvm/gems/ruby-2.7.5@ucrate/gems/activerecord-5.2.4.6/lib/active_record/connection_adapters/sqlite3_adapter.rb:32: warning: rb_check_safe_obj will be removed in Ruby 3.0

scherztc:
Top level ::CompositeIO is deprecated, require 'multipart/post' and use `Multipart::Post::CompositeReadIO` instead!
Top level ::Parts is deprecated, require 'multipart/post' and use `Multipart::Post::Parts` instead!
/Users/scherztc/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/activerecord-5.2.4.6/lib/active_record/connection_adapters/sqlite3_adapter.rb:32: warning: rb_check_safe_obj will be removed in Ruby 3.0
/Users/scherztc/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/dry-schema-1.10.2/lib/dry/schema/types_merger.rb:71: warning: Pattern matching is experimental, and the behavior may change in future versions of Ruby!
/Users/scherztc/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/activerecord-5.2.4.6/lib/active_record/connection_adapters/sqlite3_adapter.rb:32: warning: rb_check_safe_obj will be removed in Ruby 3.0

Janell-Huyck:
The fix for this issue is in 2 separate PR's.  The first, PR #1123, creates a rake task for the developers to run locally to prepare their local sqlite databases for the update.  The second, PR #1124, modifies the application.rb file to prevent future saving of data in the deprecated format.  All developers need to run the rake tasks from #1123 before they attempt to run the app with the configuration change in #1124.

Janell-Huyck:
Issues now handled by Jira.

PR #1123, opened 2023-11-13
985 - Add new rake task to convert boolean values

Janell-Huyck:
Ref #985 

#### Overview

We're addressing a deprecation warning in our application, originating from Rails 5.2.4.6, concerning the storage of boolean values in SQLite databases. This update is a proactive step towards our planned migration to Rails 6.0.

#### The Warning and Its Implications

In our current setup, boolean values in SQLite databases (used in our development and testing environments) are stored as 't' (true) and 'f' (false). However, Rails 6.0 expects these values to be stored as integers: 1 (true) and 0 (false). Although our current Rails version can work with 1/0, our SQLite databases predominantly use the 't'/'f' format.

#### Why This Update Is Necessary

Updating these values to 1/0 is crucial for compatibility with Rails 6.0. Without this change, we risk encountering boolean-related issues post-migration, as the new Rails version will not interpret 't'/'f' as boolean values.

#### Introducing the Rake Task

To facilitate this transition, we've developed a Rake task that converts all boolean fields from 't'/'f' to 1/0 in our SQLite databases. This task ensures a smooth upgrade path to Rails 6.0.

#### Running the Task: Detailed Instructions

To run this task in your local (development) and testing environments, use `bundle exec` to ensure the task executes within the context of our app's current bundle. Here are the commands for both environments:

- **Development Environment:**
  ```bash
  bundle exec rake convert_boolean_values
  ```
  This command will run the task in your development environment.

- **Testing Environment:**
  ```bash
  RAILS_ENV=test bundle exec rake convert_boolean_values
  ```
  This command explicitly sets the environment to 'test' before running the task.

- **Before testing that it works**
Completely restart the app, including restarting all background processes like Fedora, Solr, and Redis.  Failure to do this can lead to unexpected errors when running locally.  (Go ahead and ask me how I know... and what those new bald spots are!)

#### Scope and Future Considerations

This task is specifically designed for SQLite databases and does not affect our MySQL environments (QA and production). Looking ahead, aligning our development and testing setups with QA and production by using MySQL might be beneficial for consistent environment management.

#### Conclusion

Implementing this update is a key part of ensuring our application remains compatible and stable with the upcoming Rails 6.0 upgrade. It reflects our commitment to maintaining a robust and up-to-date codebase.

### DIFF
--- a/lib/tasks/convert_boolean_values.rake
+++ b/lib/tasks/convert_boolean_values.rake
@@ -18,10 +18,10 @@
 # in a development or staging environment before executing it in a production setting.
 #
 # To run it for your testing environment:
-# RAILS_ENV=test bundle exec rake db:convert_boolean_values
+# RAILS_ENV=test bundle exec rake convert_boolean_values
 #
 # To run it for your development environment:
-# bundle exec rake db:convert_boolean_values
+# bundle exec rake convert_boolean_values
 #
 # Important Notes:
 # - Ensure a database backup is taken before running this task in production.
@@ -33,6 +33,9 @@ task convert_boolean_values: :environment do
   ActiveRecord::Base.descendants.each do |model|
     # Wrap updates for each model in a transaction
     model.transaction do
+      # Skip models without a database table
+      next unless model.table_exists?
+
       puts "Updating boolean columns for #{model.name}"
 
       # Iterate over each column in the model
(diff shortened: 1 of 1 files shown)


Janell-Huyck:
We decided not to go with this approach but rather have developers drop their local tables when we switch to Rails 6.

PR #1163, opened 2024-05-02
Sets version for bunlder on capistrano deploy.

### LINKED ISSUE #979: Specify bundler version when deploying through Capistrano
See : gem install --user-install bundler:1.17.3 as Scholar user.

Might need to modify config/deploy.rb

scherztc:
Fixes #979 

The capistrano process needs a version of bundler that works with our version of ruby.


hortongn (changes requested):
@scherztc If we are pinning the bundler version used on the server, I think we should also pin the bundler gem in our Gemfile to the same 2.4.22 version.  That way we'll use the same bundler version locally as the servers.  See https://github.com/uclibs/ucrate/blob/develop/Gemfile#L58

hortongn (approved):

hortongn (changes requested):
@scherztc CircleCI is failing.  The bundler version needs to be updated here: https://github.com/uclibs/ucrate/blob/develop/.circleci/config.yml#L14

hortongn (approved):

PR #1164, opened 2024-05-08
Updates embargo scripts.

### LINKED ISSUE #1165: Date utility is broken outside of Mac environments
### Descriptive summary

The date -v breaks outside of the mac environments.

These scripts need to be cleaned up so that we can adjust crontab to use the scripts instead of directly in crontab.

### Expected behavior

cd /srv/apps/curate_uc
export PATH=$PATH:/srv/apps/.gem/ruby/2.7.0/bin
export RELEASE_DATE=`date +\%Y-\%m-\%d -d "+1 day"`
RAILS_ENV=production bundle exec rake embargo_release["$RELEASE_DATE"]

### Actual behavior

cd "$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )" )"
export RELEASE_DATE=`date -v +1d "+%Y-\%m-\%d"`
bundle exec rake embargo_release["$RELEASE_DATE"]

scherztc:
Fixes #1165 

The embargo scripts are not being used in cron since the date -v function is crashing.  We need to clean up these scripts and then set them in crontab.  

We have also separated the notification and releasing into their own scripts instead of doing it crontab.

### DIFF
--- a/script/embargo_notify.sh
+++ b/script/embargo_notify.sh
@@ -1,3 +1,29 @@
-cd /srv/apps/curate_uc
-export PATH=$PATH:/srv/apps/.gem/ruby/2.7.0/bin
-RAILS_ENV=production bundle exec rake embargo_notify
+#!/bin/sh
+
+# Notify editors that a work has been released
+# script/embargo_notify.sh [production|development]
+
+ENVIRONMENT=$1
+
+APP_DIRECTORY="$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )" )"
+
+
+if [ $# -eq 0 ]; then
+    echo -e "ERROR: no environment argument [production|development] provided"
+    exit 1
+fi
+
+if [ $ENVIRONMENT != "production" ] && [ $ENVIRONMENT != "development" ]; then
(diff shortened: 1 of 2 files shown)


hortongn (commented):
@scherztc Hard coding RAILS_ENV=production means these scripts won't work in development.  Is that ok?  Thoughts?

scherztc:
Yeah.  I had concerns.  
I didn't see environment in the .env files. 
So, I am working to find a way to pull that information from the environment.
Any ideas?

hortongn:
Perhaps the same way we handle script/restart_sidekiq.sh.  We call that script with a parameter to specify the right environment.  Like `script/restart_sidekiq.sh development` or `script/restart_sidekiq.sh production`

Look at the script/restart_sidekiq.sh source for how we're using ENVIRONMENT to grab and use the parameter.

hortongn (changes requested):
@scherztc You can remove the `THREADS=$2` and `RVM=$3` since we're not using those variables.  Also remove the banner lines.  We want cron tasks to run without output unless there's an error.  Cron sends emails whenever there's output from a cron job.

scherztc:
I tested this on DEV and it worked successfully with the application in INLINE mode verse SIDEKIQ mode.
I repeat these scripts works on DEV

hortongn (approved):

PR #1170, opened 2025-01-14
LIBSCHOLAR-41 Replaces read_timout with timout in Solr initializer.

### LINKED ISSUE #986: Deprecation Warning : Rsolr 3
DEPRECATION: Rsolr.new/connect option `read_timeout` is deprecated and will be removed in Rsolr 3. `timeout` is currently a synonym, use that instead.

How to re-create : bundle exec rake spec

Done looks like :  No warnings

scherztc:
Fixes #986

The embargo cron jobs were spamming email because of the RSolr DEPRECATION warning.  This patch to the solr_service.rb fixes the DEPRECATION removing the message and the CRON notification

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,6 +64,7 @@ Metrics/MethodLength:
     - 'app/helpers/mark_helper.rb'
     - 'bin/importer'
     - 'spec/controllers/bulkrax/importers_controller_spec.rb'
+    - 'lib/active_fedora/solr_service.rb'
 
 Metrics/ClassLength:
   Exclude:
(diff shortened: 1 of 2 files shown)


Janell-Huyck:
✅ Approved

Thanks for the clear fix to the rsolr deprecation warning.

This monkey patch overrides ActiveFedora::SolrService to replace the deprecated read_timeout option with timeout, which is now the preferred and future-proof configuration.

Since we're pinned to active-fedora ~> 12.2.4 and still on Ruby 2.x, upgrading to rsolr 3.x (which requires Ruby ≥ 3.1) isn't feasible yet. This patch is a clean interim solution that silences the warning without requiring an upstream gem change or broader dependency updates.

I've confirmed that this file will override the gem version as intended, and the RuboCop exclusion is appropriate given the legacy method lengths. Once we're in a position to upgrade active-fedora and Ruby, we can safely remove this override.

Nice work!

I added an explanatory header comment to to the top of the `lib/active_fedora/solr_service.rb` file.

Janell-Huyck (approved):

scherztc:
+1

PR #1178, opened 2025-03-05
LIBSCHOLAR-16 Bundle Update PR

scherztc:
Fixes LIBSCHOLAR-16.

Bundle updates high bundler audit vulnerabilities around redis, willow_sword, and sidekiq

With the redis upgrade, there needed to be a monkey patch to handle the redis.current call that is no longer supported.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,6 @@ gem 'hydra-remote_identifier', github: 'uclibs/hydra-remote_identifier', branch:
 gem 'kaltura', '0.1.1'
 gem 'rack', '2.2.3'
 gem 'sidekiq-limit_fetch'
-gem 'willow_sword', github: 'notch8/willow_sword'
 
 # Hyrax improvements
 gem "okcomputer", "~> 1.18.4"
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ gem 'turbolinks', '~> 5'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.5'
 # Use Redis adapter to run Action Cable in production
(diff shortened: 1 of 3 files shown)


Janell-Huyck:
We ran into several environment-specific issues while setting up Scholar@UC (ucrate) on macOS with Apple Silicon (M1/M2) and Ruby 2.7.8, and resolved them with the following changes:

## ✅ Summary of Fixes Applied

### Resolved missing libmysqlclient.22.dylib error:
- The project depends on the mysql2 gem, which by default expected a MySQL 5.7/8 client library that was no longer available.
- Our system had MySQL 9.0+, which was incompatible with mysql2 0.4.10.
- Upgraded mysql2 gem to 0.5.6:
- Updated the Gemfile to use a newer version of mysql2 that supports MySQL 9.
- Replaced:
`gem 'mysql2', '~> 0.4.10'` with: `gem 'mysql2', '~> 0.5.6'`
- Manually configured linker flags for missing libraries:
- Native gem compilation failed because the linker couldn’t find zstd, a dependency of MySQL 9.
- Fixed by explicitly configuring bundler with custom flags:
```
bundle config build.mysql2 \
  "--with-ldflags='-L/opt/homebrew/opt/mysql/lib -L/opt/homebrew/opt/zstd/lib' \
   --with-cppflags='-I/opt/homebrew/opt/mysql/include -I/opt/homebrew/opt/zstd/include'"
```
- Successfully reinstalled gems and ran migrations:

After applying the above changes, bundle install and bundle exec rake db:migrate ran successfully.
Some non-blocking deprecation warnings were noted (e.g., CompositeIO, rb_check_safe_obj), but did not prevent migration.

Janell-Huyck (approved):

PR #1184, opened 2025-08-14
LIBSCHOLAR-53 : Updates Google Analytics ID to version 4.

scherztc:
Fixes LIBSCHOLAR-53

Use Environment variable for ID.
Overwrite Hyrax GA Partial.
This only fixes tracking the visits in Google Analytics, but we still may have to do more for Scholar statistics to work fully.

Make sure to set .env variable with new ID.


haitzlm (approved):
Comments:
-Verified correct account on scholar-dev
-Verified data collection on scholar-dev

PR #1196, opened 2026-03-23
LIBSCH1-67 Update bundler-audit.yml

Janell-Huyck:
Adds new vulnerabilities to bundler-audit skip list


haitzlm (approved):
OK, simple enough

PR #1199, opened 2026-04-09
Css updates

haitzlm:
Accessibility Fix.  Buttons on home page contrast error when checked with WAVE.

FIx: Updated CSS for buttons on home page header image to solid background. Also ignore ide (ruby Mine) project files /idea


Janell-Huyck:
Looks good to merge when CI is complete

Janell-Huyck (approved):

PR #1208, opened 2026-04-17
LIBSCH1-76 Bug fix: menu width correction

Janell-Huyck:
Fix sidebar facet menu width when body uses flex layout

This PR fixes an issue where the sidebar facet menu items appeared only one character wide after the body was set to `display: flex`. The solution is to explicitly set `display: block` on the facet menu container and its children, ensuring correct width and layout.

**Changes:**
- Add `display: block` to `.facet-values`, `.facet-values > *`, `.facet-label`, and `.facet-count` in `hyrax/dashboard.scss`.
- Add `width: 100%` to the same to allow labels to display in one line

No other changes are included. This resolves the menu width regression while preserving the site-wide flexbox layout.

Before this change:
Screenshot 2026-04-17 at 9 28 30 AM

After this change:
Screenshot 2026-04-17 at 11 50 27 AM
Screenshot 2026-04-17 at 11 50 21 AM

### DIFF
--- a/app/assets/stylesheets/hyrax/dashboard.scss
+++ b/app/assets/stylesheets/hyrax/dashboard.scss
@@ -1,3 +1,10 @@
+// Ensure facet menu and its children are not flex items
+.facet-values,
+.facet-values > *,
+.facet-label,
+.facet-count {
+  display: block;
+}
 @import 'morris.js/0.5.1/morris';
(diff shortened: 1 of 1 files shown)


haitzlm (approved):
The CSS is correct. Though I dont have this running locally, at least i can confirm the CSS.

